PR #2, opened 2015-11-21
(QA-2081) Gemify rototiller

zreichert:


er0ck:
this could probably use a rubocop file also, but we can (maint) that later if it doesn't make it in before i build this thing

Rakefile
```diff
@@ -0,0 +1,12 @@
+require 'rototiller'
+require "bundler/gem_tasks"
```
er0ck, line 1: why are these two requires here?  doesn't appear they are needed?

lib/rototiller/utilities/env_var_checker.rb
```diff
@@ -24,7 +24,7 @@ def check_env_vars
     end
     abort_message = 'Aborting Rake:'
     required_vars.each{ |v| abort_message << red_text("\nThe environment variable #{v.name} is required. #{v.message}")}
-    abort(abort_message) unless required_vars.empty?
+    raise abort_message unless required_vars.empty?
```
er0ck: this should probably use an exception class like 
http://ruby-doc.org/core-2.2.0/ArgumentError.html

it should probably also be handled (read: logged, styled) in an exception handler upstream.  but i'm inclined to let this one in for now because we don't have a bunch of these messages, yet.

Rakefile
zreichert, line 1: @er0ck The 'bundler/gem_tasks' require adds tasks for 'build', 'install', and 'release'.  I am not sure but I believe that the Gem pipelines in Jenkins may use it.

lib/rototiller/utilities/env_var_checker.rb
zreichert: @er0ck  Do you think I should implement a Rototiller::RototillerError class?

er0ck: yeah, but maybe ticket it for later?

Gemfile
```diff
@@ -0,0 +1,3 @@
+source 'https://rubygems.org'
+# Specify your gem's dependencies in scooter.gemspec
```
cowofevil: Scooter?

Rakefile
```diff
@@ -0,0 +1,11 @@
...
+
```
cowofevil: Superfluous newline.

spec/rototiller/env_var_checker_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+  end
+
+  context 'Checking with and without defaults' do
+    
+  end
+
+  context 'With and without defaults' do
+
+  end
```
cowofevil, line 69: Create tickets to do this work and make a TODO comment.

rototiller.gemspec
```diff
@@ -0,0 +1,30 @@
...
+  s.description   = 'Puppetlabs tool for building rake tasks'
+  s.homepage      = "https://github.com/puppetlabs/rototiller"
+  s.version       = Rototiller::Version::STRING
+  s.license       = 'Apache-2.0'
+  s.files         = Dir['[A-Z]*[^~]'] + Dir['lib/**/*.rb'] + Dir['spec/*']
+
+  #Development dependencies
+  s.add_development_dependency 'bundler', '~> 1.6'
+  s.add_development_dependency 'pry', '~> 0.9.12'
+  s.add_development_dependency 'rake'
```
cowofevil: Duplicate dependency with rake. This confuses bundler.

PR #3, opened 2015-12-16
(QA-2078) New beaker task

zreichert:


zreichert:
@er0ck For a demo look at the Rakefile provided in demo/

lib/rototiller/rake/dsl/dsl_extention.rb
```diff
@@ -1,14 +1,11 @@
...
 module Rake
   module DSL
 
     def acceptance_task(*args, &block)
-      Rake::RototillerTask.define_task :acceptance, &block
-      description = "Tests in the 'Acceptance' tier"
-      unless Rake::RototillerTask[:acceptance].comment
-        Rake::RototillerTask[:acceptance].add_description(description)
-      end
+      desc "Tests in the 'Acceptance' tier" unless ::Rake.application.last_comment
```
er0ck, line 9: please comment this with why there is a conditional.  something like "set a description for this task unless the user has already done so"

lib/rototiller/task/beaker_task.rb
```diff
@@ -0,0 +1,74 @@
...
+        @verbose       = true
+        @fail_on_error = true
+
+        define(args, &task_block)
+      end
+
+      def self.define_task(*args, &task_block)
+        self.new(*args, &task_block)
+      end
+
```
er0ck, line 31: please comment this to talk about why we want define_task as an "alias" to init()

lib/rototiller/task/beaker_task.rb
```diff
@@ -0,0 +1,74 @@
+require 'rake'
+require 'rototiller/task/flags/cli_flags'
+
+module Rototiller
+  module Task
+    class BeakerTask
```
er0ck, line 6: this seems like it could get most of its functionality from RototillerTask.  should it inherit?

as discussed with Zach, maybe later.

lib/rototiller/task/beaker_task.rb
```diff
@@ -0,0 +1,74 @@
...
+        puts command if verbose
+
+        # currently this does not execute
+        # as a Demo this should not actually run beaker
+      end
+
+      private
+
+      def define(args, &task_block)
+        desc "A Beaker Task" unless ::Rake.application.last_comment
```
er0ck: this description should be customizable via instance setters

lib/rototiller/task/beaker_task.rb
```diff
@@ -0,0 +1,74 @@
...
+        task(@name, *args) do |_, task_args|
+          RakeFileUtils.__send__(:verbose, verbose) do
+            task_block.call(*[self, task_args].slice(0, task_block.arity)) if task_block
+            run_task(verbose)
+          end
+        end
+      end
+
+      def beaker_command
+        # should we bundle here ?????
```
er0ck: i don't think so.  we typically bundle exec from outside of rake.  
machines without root, all my stuff is in my bundlepath.  i believe that's how jenkins works also.  we're always running from bundler.
in addition to that, i often run with global gem installs from either rake or beaker with no bundle.
so maybe later we can have an option to this class to use bundler.  or we end up with a beakerCommand class that builds the command and we delegate in the Task Class to a beaker/Command class...

lib/rototiller/task/beaker_task.rb
```diff
@@ -0,0 +1,74 @@
...
+        beaker += " --config #{@helper}" if @helper
+        beaker += " --type #{@type}" if @type
+        beaker += " --keyfile #{@keyfile}" if @keyfile
+        beaker += " --options-file #{@options_file}" if @options_file
+        beaker += " --load-path #{@load_path}" if @load_path
+        beaker += " --pre-suite #{@pre_suite}" if @pre_suite
+        beaker += " --post-suite #{@post_suite}" if @post_suite
+        beaker += " --tests #{@tests}" if @tests
+        #beaker += " #{ENV["BEAKER_OPTS"]}" if o
+        return beaker
```
er0ck, line 71: this has a pretty high branch complexity.  i'll PR in a rubocop file :-) and maybe a way to abstract this a bit so we don't have the repetition with the CLIFlags module

PR #6, opened 2016-01-15
(QA-2329) refactor EnvVar class

zreichert:
I'll move this file to a more appropriate location before merge.


er0ck:
amend that other commit with (maint) or (ticketnum)

lib/rototiller/utilities/env_var_checker.rb
```diff
@@ -1,40 +1,40 @@
...
+  def initialize(var, message, default=false)
+
+    @var = var
+    @message = message
+    @default = default
+    @value = ENV[@var] || @default
   end
 
-  def track_env_var(var, message, required=false)
-    @@vars.push(@@env.new(var, message, required))
```
er0ck, line 37: exactly.  i think we can move to RototillerTask having the list, we push that list to dependent tasks.

lib/rototiller/utilities/env_var_checker.rb
```diff
@@ -1,40 +1,40 @@
...
+      red_text("The ENV #{@var} is required, #{@message}")
+    elsif !@default && check
+      # ENV is present and it has no default value
+      green_text("The ENV #{@var} was found in the environment with the value #{@value}")
+    elsif @default && !check
+      # ENV is not present and it has default value
+      yellow = yellow_text("WARNING: the ENV #{@var} is not set, proceeding with default value: ")
+      green = green_text(@value)
+      yellow << green
+    end
```
er0ck: hmmm... this is a bit smelly.  not sure how i'd fix it though...
maybe a method to determine its defaultness and checkness then case/elsif tree on that?

i guess it's parametrically constrainied on the two and all 4 cases are accounted :neutral_face:

er0ck:
and this will obviously need some tests.  maybe rename the current EnvVar tracking module so existing rototillerTask and acceptanceTask will still work, if possible.  then we can move/scrap those for the ticket in current sprint?

lib/rototiller/utilities/env_var_checker.rb
zreichert: Yeah this is not my favorite either...

spec/rototiller/utilities/env_var_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    context method_signature do
+
+      ['ENV set', 'ENV not set'].each do |env_set|
+
+        context env_set do
+
+          let (:var_name)       { "VARNAME_#{(0...8).map { (65 + rand(26)).chr }.join}" }
+          let (:var_message)    { "This is how you use #{var_name}" }
+          let (:var_default)    { "VARDEFAULT_#{(0...8).map { (65 + rand(26)).chr }.join}" }
+          let (:var_env_value)  { "VARENVVALUE_#{(0...8).map { (65 + rand(26)).chr }.join}" }
```
er0ck, line 18: this is cool.  i should use this randomization for task names in RototillerTask tests

lib/rototiller/utilities/env_var.rb
```diff
@@ -0,0 +1,55 @@
...
+    # @default_specified = ????
+    @default = default
+    #test = ENV[var]
+    #binding.pry
+    #if test
+    #  @value = test
+    #else
+    #  @value = @default
+    #end
+    #binding.pry
```
er0ck: remove the above

lib/rototiller/utilities/env_var.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  include ColorText
+
+  attr_accessor :var, :message, :default
+
+  def initialize(var, message, default = nil)
+    # require 'pry'; binding.pry
+    @var = var
+    @message = message
+    # TODO: maybe see if the value was passed in the args array or not as a default (so that nil may be passed)
```
er0ck: i don't understand why you'd want to pass nil.  but either way default will just get set to nil, which is probably appropriate, and creates a required envvar, as far as i can tell...  tests seem to confirm this.

lib/rototiller/utilities/env_var.rb
```diff
@@ -0,0 +1,55 @@
...
+    elsif @default && !check
+      # ENV is not present and it has default value
+      yellow = yellow_text("WARNING: the ENV #{@var} is not set, proceeding with default value: ")
+      green = green_text(value)
+      yellow << green
+    end
+  end
+
+  private
+  def check
```
er0ck, line 39: should this set a status instance var, and should it check if there is default?
maybe check is just for existence of ENV and something like is_valid?() (public) would tell if the var is usable?  then the collector would know if to spew the warnings/errors.
i'd be okay doing that when we implement the collector

lib/rototiller/utilities/env_var.rb
zreichert: This was a comment left by Ethan.

PR #8, opened 2016-01-29
(QA-2361) add collector class

zreichert:


lib/rototiller/utilities/collector.rb
```diff
@@ -0,0 +1,65 @@
+require 'rototiller/utilities/flag'
+require 'rototiller/utilities/env_var'
+require 'forwardable'
+
+class ParamCollection
+
+  # This may be useful if we use individual ParamCollection objects for EnvVars and Flags
+  extend Forwardable
```
er0ck: :+1: this is probably better for things like Forwardable rather than <, which i've used in the past since this isn't _a_ forwardable.  :neutral_face:

lib/rototiller/utilities/env_var.rb
```diff
@@ -5,11 +5,13 @@ class EnvVar
   include ColorText
 
   attr_accessor :var, :message, :default
+  attr_reader :message_level, :stop
```
er0ck, line 8: better names!

spec/rototiller/utilities/collector_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      [
+          set_env_1_no_default, set_env_2_no_default
+      ].each do |var|
+        expected_message = /32mThe ENV #{var.var} was found in the environment with the value #{var.value}/
+        expect(param_collection.format_messages({:default => false, :message_level => :info})).to match(expected_message)
+      end
+    end
+  end
+
+end
```
er0ck: should this class display the env's when --verbose is used?  maybe not important now

spec/spec_helper.rb
```diff
@@ -13,6 +13,29 @@ def random_string
...
+  begin
+    old_stdout = $stdout
+    $stdout = StringIO.new('','w')
+    yield
+    $stdout.string
+  ensure
+    $stdout = old_stdout
+  end
+end
+
```
er0ck, line 38: nice

er0ck:
looks good, dude.  what else is in the works before this is ready to go?

zreichert:
@er0ck I think I'm ready for a final review.

er0ck:
two more failures in rototiller_task.  but your tests run fine.  so i'm inclined to merge

PR #15, opened 2016-03-21
(QA-2229) First acceptance tests for rototiller

zreichert:


Rakefile
```diff
@@ -8,3 +8,52 @@ RSpec::Core::RakeTask.new(:test) do |t|
...
+  beaker += " --config #{config}" if config != ''
+  beaker += " --type #{type}" if type != ''
+  beaker += " --keyfile #{keyfile}" if keyfile != ''
+  beaker += " --load-path #{load_path}" if load_path != ''
+  beaker += " --pre-suite #{pre_suite}" if pre_suite != ''
+  beaker += " --post-suite #{post_suite}" if post_suite != ''
+  beaker += " --tests #{test_suite}" if test_suite != ''
+  beaker += " #{opts}" if opts != ''
+  sh beaker
+end
```
er0ck: wow, i can't wait to dogfood our own stuff!

Rakefile
```diff
@@ -8,3 +8,52 @@ RSpec::Core::RakeTask.new(:test) do |t|
...
 end
+
+task :generate_host_config do |t, args|
+  if ENV["BEAKER_CONFIG"]
+    next
+  end
+  if not ENV['TEST_TARGET']
+    fail "FATAL: must set TEST_TARGET environment variable."
+  end
+
```
er0ck, line 16: we don't want a default for this?  
orch:
`ENV['TEST_TARGET'] || 'debian7-64af-64af-64af-64af'`

rototiller.gemspec
```diff
@@ -17,6 +17,8 @@ Gem::Specification.new do |s|
   #Development dependencies
   s.add_development_dependency 'rspec', '~> 3.1.0'
   s.add_development_dependency 'simplecov'
+  s.add_development_dependency 'beaker'
```
er0ck: this is in the :system_tests group in the Gemfile so it can be excluded upon bundle install if one wants. 
do we think it's a development dependency?  maybe
whatever we decide, same should be true of hostgen, i suppose

PR #16, opened 2016-03-23
(QA-2406) update method signatures for EnvVar.new and CommandFlag.new

zreichert:


er0ck:
ruby 1.9.3 unit test failure not associated with change

PR #22, opened 2016-03-29
Fix ENVs not being set, wow thats embarrassing

zreichert:


lib/rototiller/version.rb
```diff
@@ -1,5 +1,5 @@
 module Rototiller
   module Version
-    STRING = '0.0.0'
+    STRING = '0.0.0.b'
```
er0ck: separate commit, please

lib/rototiller/utilities/env_var.rb
```diff
@@ -16,17 +16,12 @@ class EnvVar
   # @param var [String] the ENV in the environment, 'HOME'
   # @param message [String] the message describing the ENV
   # @param default [String] the value to use as the default if the ENV is not present
-  def initialize(var, default=false, message)
+  def initialize(var, default=nil, message)
```
er0ck: :+1: we might want this in a separate commit as well.

er0ck:
this needs a rebase, please

PR #23, opened 2016-03-31
(QA-2429) Dogfood rototiller into rototillers rake task

zreichert:


Rakefile
```diff
@@ -22,21 +22,21 @@ task :generate_host_config do |t, args|
...
 
-task acceptance: :generate_host_config
+#task acceptance: :generate_host_config
 
 desc 'Run acceptance tests for Rototiller'
-task :acceptance do |t, args|
+task :acceptance => [:check_acceptance, :generate_host_config]do |t, args|
 
-  config = ENV["BEAKER_CONFIG"] || 'acceptance/hosts.cfg'
+  config = ENV["BEAKER_CONFIG"] #|| 'acceptance/hosts.cfg'
```
er0ck: nit: remove comment
and below.  the defaults are retained in the check_blah tasks

Rakefile
```diff
@@ -1,18 +1,18 @@
...
 require 'rspec/core/rake_task'
+require 'rototiller'
 
 task :default => :test
 
 desc "Run spec tests"
 RSpec::Core::RakeTask.new(:test) do |t|
   t.rspec_opts = ['--color']
-  t.pattern = 'spec/'
+  t.pattern = ENV['SPEC_PATTERN']
```
er0ck, line 10: i _seeee_ 
yeah, in this one we need a separate task to check because we're not using a rototiller task here, it's an rspec RakeTask, so we can't futz with the command using EnvVars.  
but i think in our acceptance task we could?  maybe having them in separate tasks is best practice?  

can you spell out these two cases in the README/docs?

Rakefile
```diff
@@ -49,8 +49,22 @@ task :acceptance do |t, args|
...
 end
+
+Rototiller::Task::RototillerTask.define_task :check_acceptance do |t|
+  t.add_env('TEST_TARGET', 'centos7-64', 'The argument to pass to beaker-hostgenerator')
+  t.add_env('BEAKER_CONFIG', 'acceptance/hosts.cfg', 'The configuration file that Beaker will use')
+  t.add_env('BEAKER_PRESERVEHOSTS', 'never', 'The beaker setting to preserve a provisioned host')
+  t.add_env('BEAKER_KEYFILE', "#{ENV['HOME']}/.ssh/id_rsa-acceptance", 'The SSH key used to access a SUT')
+  t.add_env("BEAKER_LOADPATH", 'acceptance/lib', 'The load path Beaker will use')
+  t.add_env("BEAKER_PRESUITE", 'acceptance/pre-suite', 'THe path to a directory containing pre-suites')
+  t.add_env("BEAKER_TESTSUITE", 'acceptance/tests', 'The path to the tests you want beaker to run')
```
er0ck: nit: might want to align on commas

Rakefile
zreichert, line 10: @er0ck I think this will change once we introduce the new block syntax.

PR #24, opened 2016-03-31
(WIP) example of new block syntax

zreichert:
Checkout the 'block_syntax' task on the Rakefile.


er0ck:
this is fantastic. i think we'll have to do this for add_flag as well, to keep the API consistent

PR #25, opened 2016-04-07
(qa-2448) POC; DO NOT MERGE increase rototiller acceptance coverage

er0ck:
@zreichert let me know what you think of the refactor.  i'll add tests in other commits later


acceptance/tests/rototillerTask/environment_variables.rb
```diff
@@ -0,0 +1,61 @@
...
+      assert(result.exit_code == 0, 'The expected exit code 0 was not observed')
+      assert_no_match(/error/i, result.output, 'An unexpected error was observed')
+
+      env_vars.each do |env|
+        # validate notification to user of ENV value
+        #assert_match(/, result.stdout, 'The expected messaging was not observed')
+
+        # Use test command output to validate value of ENV used by task
+        #assert_match(command_regex, result.stdout, 'The observed value of the ENV was different than expected')
+      end
```
er0ck, line 57: zreichert note these aren't turned on yet, nor finished.  
i believe this is catching a bug.  the no default, no exist case should exit, but it appears it's being masked by the later add_env() case?  (it's producing the error message, but without ERROR)

```
        present value
        present value

        notpresent default value

        The ENV NO_DEFAULT-EXISTS was found in the environment with the value present value
        The ENV DEFAULT-EXISTS was found in the environment with the value present value
        The ENV NO_DEFAULT-NO_EXISTS is required, no default, does not previously exist
        WARNING: the ENV DEFAULT-NO_EXISTS is not set, proceeding with default value: notpresent default value
```

possibly also a problem where ERROR/INFO isn't being printed in some cases

i need to refactor the naming of the cases in here so they are more clear/concise

i can probably patch up the problems tomorrow

PR #27, opened 2016-04-08
(QA-2447) Equivalent env

zreichert:


Rakefile
```diff
@@ -22,74 +22,51 @@ task :generate_host_config => [:check_acceptance]do |t, args|
...
     env.message = 'The configuration file that Beaker will use'
+    env.override_env = 'BEAKER_CONFIG'
   end
-  t.add_env do |env|
-    env.name = 'BEAKER_PRESERVEHOSTS'
+  t.add_flag do |env|
+    env.name = '--preserve-hosts'
     env.default = 'never'
     env.message = 'The beaker setting to preserve a provisioned host'
+    env.override_env = 'BEAKER_PRESERVEHOSTS'
```
er0ck: i think it might be good to have same spelling of envs and flags e.g.: `BEAKER_PRESERVE-HOSTS`

Rakefile
```diff
@@ -22,74 +22,51 @@ task :generate_host_config => [:check_acceptance]do |t, args|
...
-    env.name = 'BEAKER_CONFIG'
+  t.add_flag do |env|
+    env.name = '--hosts'
     env.default = 'acceptance/hosts.cfg'
     env.message = 'The configuration file that Beaker will use'
+    env.override_env = 'BEAKER_CONFIG'
   end
-  t.add_env do |env|
-    env.name = 'BEAKER_PRESERVEHOSTS'
+  t.add_flag do |env|
```
er0ck: `do |flag|`

Rakefile
```diff
@@ -12,7 +12,7 @@ end
 
 task :test => [:check_test]
 
-task :generate_host_config => [:check_acceptance]do |t, args|
+task :generate_host_config do |t, args| #=> [:check_acceptance]do |t, args|
 
   target = ENV["TEST_TARGET"] || 'centos7-64'
   generate = "bundle exec beaker-hostgenerator"
```
er0ck: until we have add_arg() or whatever, we could use a flag and ENV for #{target} below.  and command.  oh, and we shouldn't need to bundle exec here

zreichert: @er0ck I think we will have to wait for host generator, since it directs the output to a file.  We could do that with a boolean flag but that feels hackish.

Rakefile
```diff
@@ -12,76 +12,61 @@ end
 
 task :test => [:check_test]
 
-task :generate_host_config => [:check_acceptance]do |t, args|
+task :generate_host_config do |t, args| #=> [:check_acceptance]do |t, args|
```
er0ck: remove comment, pls

Rakefile
```diff
@@ -12,76 +12,61 @@ end
...
+    flag.message = 'The configuration file that Beaker will use'
+    flag.override_env = 'BEAKER_HOSTS'
   end
-  t.add_env do |env|
-    env.name = 'BEAKER_PRESERVEHOSTS'
-    env.default = 'never'
-    env.message = 'The beaker setting to preserve a provisioned host'
+  t.add_flag do |flag|
+    flag.name = '--preserve-hosts'
+    flag.default = 'never'
```
er0ck: nit: actual functionality of this file.  should this default to onfail?

zreichert: I would think this would be used in CI, I didn't think that 'onfail' was appropriate in CI.

er0ck: i think all the puppet tests are setup that way.  so you can go later(-ish) and login to the machine to see what's up

Rakefile
```diff
@@ -12,76 +12,61 @@ end
...
-  beaker += " --hosts #{config}" if config != ''
-  beaker += " --keyfile #{keyfile}" if keyfile != ''
-  beaker += " --load-path #{load_path}" if load_path != ''
-  beaker += " --pre-suite #{pre_suite}" if pre_suite != ''
-  beaker += " --tests #{test_suite}" if test_suite != ''
-  sh beaker
-end
-
-Rototiller::Task::RototillerTask.define_task :check_acceptance do |t|
+Rototiller::Task::RototillerTask.define_task :acceptance => [:generate_host_config] do |t|
```
er0ck, line 25: no need to hashrocket to an array here, but i dunno if i care :neutral_face:

PR #35, opened 2016-04-20
support overriding with command line arguments

samwoods1:
Not ready for merge, just wanted to check to see if this is what you guys wanted.  If so, I can add some unit tests.


er0ck:
@samwoods1 did you still want to get this one somewhere?

we should probably talk about this as a possible UI for users when we refactor the API

samwoods1:
Meh...  I didn't really get how to set environment variables on the command line when calling rake.  Now that I know how to do that, this isn't really important.

PR #36, opened 2016-04-21
(QA-2457) override_env for command

zreichert:
We need to decide how we want to collect information about the command to be executed. 

```
t.add_command( {:command => 'echo', :override_env => 'ECHO_OVERRIDE'})
```

Would :name be better than :command ? 
Do we still want to support 

```
t.command = 'echo'
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -64,9 +64,9 @@ rototiller_task :acceptance => [:generate_host_config] do |t|
     flag.override_env = 'BEAKER_TESTS'
   end
 
-  t.command = 'beaker --debug'
+  t.add_command({:name => 'beaker --debug', :override_env => 'BEAKER_EXECUTABLE'})
 end
 
 Rototiller::Task::RototillerTask.define_task :check_test do |t|
-  t.add_env(:name => 'SPEC_PATTERN', :default => 'spec/', :message => 'The pattern RSpec will use to find tests')
+  t.add_env({:name => 'SPEC_PATTERN', :default => 'spec/', :message => 'The pattern RSpec will use to find tests'})
 end
--- a/acceptance/tests/rototillerTask/override_command_with_env.rb
+++ b/acceptance/tests/rototillerTask/override_command_with_env.rb
@@ -1,0 +1,69 @@
+require 'beaker/hosts'
+require 'rakefile_tools'
+
+test_name 'C97827: can set envvar to override command name when using task.command' do
+  extend Beaker::Hosts
+  extend RakefileTools
+
+  step 'Test command with an override_env that has a value' do
+
+    override_env = 'BROCKSAMSON'
+    env_key = 'THIS_WAS_IN_ENV'
+    env_value = 'echo ' << env_key
+
+    @task_name    = 'command_flags_with_override'
+    rakefile_contents = <<-EOS
+$LOAD_PATH.unshift('/root/rototiller/lib')
+require 'rototiller'
+
+Rototiller::Task::RototillerTask.define_task :#{@task_name} do |t|
+    t.add_command({:name => 'echo', :override_env => '#{override_env}'})
+end
+    EOS
+    rakefile_path = create_rakefile_on(sut, rakefile_contents)
+    sut.add_env_var(override_env, env_value)
+
+    step 'Execute task defined in rake task' do
+      on(sut, "rake #{@task_name}", :accept_all_exit_codes => true) do |result|
+        # exit code & no error in output
+        assert(result.exit_code == 0, 'The expected exit code 0 was not observed')
+        assert_no_match(/error/i, result.output, 'An unexpected error was observed')
+
+        # command was used that was supplied by the override_env
+        assert_match(/^#{env_key}/, result.stdout, 'The correct command was not observed')
+      end
+    end
+  end
+
+  step 'Add Command with block syntax and unset override_env' do
+
+    override_env = 'EMPTYENV'
+
+    @task_name    = 'command_flags_with_override'
+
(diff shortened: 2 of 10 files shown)


er0ck:
:name is better than :command

i think we should support the same pattern we support in add_*

Rakefile
```diff
@@ -64,9 +64,9 @@ Rototiller::Task::RototillerTask.define_task :acceptance => [:generate_host_conf
     flag.override_env = 'BEAKER_TESTS'
   end
 
-  t.command = 'beaker --debug'
+  t.add_command(:command => 'beaker --debug', :override_env => 'BEAKER_EXECUTABLE')
 end
```
er0ck, line 68: doesn't this need to be a hash?

Rakefile
```diff
@@ -64,9 +64,9 @@ Rototiller::Task::RototillerTask.define_task :acceptance => [:generate_host_conf
...
 
 Rototiller::Task::RototillerTask.define_task :check_test do |t|
   t.add_env(:name => 'SPEC_PATTERN', :default => 'spec/', :message => 'The pattern RSpec will use to find tests')
-end
+end
```
er0ck: wuts up with your editor, man?

acceptance/tests/rototillerTask/override_command_with_env.rb
```diff
@@ -0,0 +1,110 @@
+require 'beaker/hosts'
+require 'rakefile_tools'
+
+test_name 'C97827: can set envvar to override command name when using task.command' do
+  extend Beaker::Hosts
+  extend RakefileTools
+
+  def create_rakefile_task_segment(flags)
```
er0ck: flags?

lib/rototiller/task/rototiller_task.rb
```diff
@@ -134,6 +160,17 @@ def add_param(collection, param_class, param_array, args, &block)
...
+      def pull_params_from_block(param_array, &block)
+
+        block_syntax_obj = Rototiller::Block_syntax.new
+        block_syntax_obj.class.class_eval { param_array.each { |i| attr_accessor i } }
+
+        yield(block_syntax_obj)
+
+        return block_syntax_obj
+
+      end
```
er0ck, line 156: should this move to BlockSyntaxObject?

lib/rototiller/utilities/block_syntax_object.rb
```diff
@@ -0,0 +1,12 @@
+module Rototiller
+  class Block_syntax
```
er0ck: maybe should remain just a module, then can be included into the Classes that use it?
feels more like a type of functionality than something to compose-in (shrug)

lib/rototiller/utilities/command.rb
```diff
@@ -0,0 +1,28 @@
...
+      if attribute_hash[:override_env]
+        @override_env = EnvVar.new({:name => attribute_hash [:override_env], :default => attribute_hash[:command]})
+        @command = @override_env.value
+      else
+        @command = attribute_hash[:command]
+      end
+    end
+  end
+end
+
```
er0ck, line 28: nice.  utilities is getting a bit crowed.  we might want to rename it for Command, Flag, EnvVar pieces

Gemfile
```diff
@@ -33,3 +33,4 @@ end
 
 # Specify your gem's dependencies in rototiller.gemspec
 gemspec
+gem 'pry'
```
er0ck: remove, please

acceptance/tests/rototillerTask/override_command_with_env.rb
```diff
@@ -0,0 +1,67 @@
...
+        assert(result.exit_code == 0, 'The expected exit code 0 was not observed')
+        assert_no_match(/error/i, result.output, 'An unexpected error was observed')
+
+        # command was used that was supplied by the override_env
+        assert_match(/^#{env_key}/, result.stdout, 'The correct command was not observed')
+      end
+    end
+  end
+
+  step 'Test command with an override_env that has a value' do
```
er0ck: rename this step.  maybe something about block syntax and empty command envvar?

acceptance/tests/rototillerTask/override_command_with_env.rb
```diff
@@ -0,0 +1,67 @@
...
+    EOS
+    rakefile_path = create_rakefile_on(sut, rakefile_contents)
+
+    step 'Execute task defined in rake task' do
+      on(sut, "rake #{@task_name}", :accept_all_exit_codes => true) do |result|
+        # exit code & no error in output
+        assert(result.exit_code == 0, 'The expected exit code 0 was not observed')
+        assert_no_match(/error/i, result.output, 'An unexpected error was observed')
+
+        # flags should be echoed to command line
```
er0ck: can remove this comment and one above.  reiterates the code, which is pretty clear.
would be :nit but the comment is wrong :-D

er0ck:
this PR should probably also rename `env_var_to_override_command_flag` to be similar to the name of this acceptance test `override_command_with_env`

acceptance/tests/rototillerTask/override_command_with_env.rb
```diff
@@ -0,0 +1,67 @@
...
+        assert_match(/^--/, result.stdout, 'The correct command was not observed')
```
er0ck: does this work?

i don't see where it would output a line starting with `--`

```
          Expected /^--/ to match "\n\n\n".>
```

should probably echo something more pertinent to assert against

er0ck:
this needs a rebase, please

zreichert:
@er0ck rebased and ready

PR #37, opened 2016-04-22
(QA-2466) add rototiller_task to the rake DSL

zreichert:


er0ck:
this needs a spec test.
something like 

```
describe 'rototiller_task' 
it 'calls the right method'
expect(subject).to receive(Rototiller::Task::RototillerTask.define_task)
```

er0ck:
acceptance passes.

PR #39, opened 2016-04-28
(QA-2446) arguments to commands

zreichert:


acceptance/tests/rototillerTask/command_arguments_with_override.rb
```diff
@@ -0,0 +1,78 @@
+require 'beaker/hosts'
+require 'rakefile_tools'
+
+test_name 'C97827: can set envvar to override command name when using task.command' do
```
er0ck: 'command args' ?
wrong case number

acceptance/tests/rototillerTask/command_arguments_with_override.rb
```diff
@@ -0,0 +1,78 @@
...
+  extend Beaker::Hosts
+  extend RakefileTools
+
+  step 'Test command with an override_env that has a value' do
+
+    override_env = 'BROCKSAMSON'
```
er0ck: probably should use a different env here than the other test.
we might want to assert we aren't using an env that already exists in these tests.

although i guess we're not really testing that env involvement here, so maybe a 😐

lib/rototiller/task/rototiller_task.rb
```diff
@@ -77,13 +78,13 @@ def add_flag(*args, &block)
       # for block {|a| ... }
       # @yield [a] Optional block syntax allows you to specify information about command, available methods track hash keys
       def add_command(args={}, &block)
-        attributes = [:name, :override_env]
+        attributes = [:name, :override_env, :argument, :argument_override_env]
```
er0ck, line 81: hmmmm.  should this not have its own method to match add_command, add_env, add_flag?
as a user i'd expect to be able to use add_arg, just like flags.
edit: decided to continue with this signature.  see how it does in beta

acceptance/tests/rototillerTask/command_arguments_with_override.rb
```diff
@@ -0,0 +1,78 @@
...
+    @task_name    = 'command_with_args_and_defaults'
+    rakefile_contents = <<-EOS
+$LOAD_PATH.unshift('/root/rototiller/lib')
+require 'rototiller'
+
+Rototiller::Task::RototillerTask.define_task :#{@task_name} do |t|
+    t.add_command({:name => 'echo', :override_env => '#{override_env}', :argument => '#{default_arg}', :argument_override_env => '#{argument_override_env}'})
+end
+    EOS
+    rakefile_path = create_rakefile_on(sut, rakefile_contents)
```
er0ck, line 29: we're starting to repeat a bunch of this stuff in tests.
perhaps create_rakefile_on should take a bunch of task block strings
the rest of this crap is implied (first three lines).
can do in another ticket.  probably should have a Rakefile class that has a bunch of task classes composed into it.
</thinking outloud>

acceptance/tests/rototillerTask/command_arguments_with_override.rb
```diff
@@ -0,0 +1,78 @@
...
+    env_value = 'echo ' << env_key
+
+    default_arg = 'HANKVENTURE'
+
+    @task_name    = 'command_with_args_and_defaults'
+    rakefile_contents = <<-EOS
+$LOAD_PATH.unshift('/root/rototiller/lib')
+require 'rototiller'
+
+Rototiller::Task::RototillerTask.define_task :#{@task_name} do |t|
```
er0ck: we should leave a few tests with this long-form.  but i think most tests should probably use `rototiller_task`, now

acceptance/tests/rototillerTask/command_arguments_with_override.rb
```diff
@@ -0,0 +1,78 @@
...
+        # command was used that was supplied by the override_env
+        assert_match(/^#{env_key} #{argument_env_key}/, result.stdout, 'The correct command was not observed')
+      end
+    end
+  end
+
+  step 'Add Command with block syntax and unset override_env' do
+
+    override_env = 'EMPTYENV'
+
```
er0ck: nit: extra blank line

acceptance/tests/rototillerTask/command_arguments_with_override.rb
```diff
@@ -0,0 +1,78 @@
...
+    @task_name    = 'command_with_args_and_defaults'
+
+    validation_string = (0...10).map { ('a'..'z').to_a[rand(26)] }.join
+    argument_validation_string = (0...10).map { ('a'..'z').to_a[rand(26)] }.join
```
er0ck: we probably need our own random_string method

lib/rototiller/task/rototiller_task.rb
```diff
@@ -132,17 +133,19 @@ def set_verbose(verbosity=true)
         @verbose = verbosity
       end
 
-      def add_param(collection, param_class, param_array, args, &block)
+      def add_param(collection, param_class, param_array, args, opts={}, &block)
 
         if block_given?
 
           param_hash = pull_params_from_block(param_array, &block).to_h
+          param_hash[:set_env] = true if opts[:set_env]
```
er0ck, line 143: yeah.  better.

lib/rototiller/task/rototiller_task.rb
```diff
@@ -132,17 +133,19 @@ def set_verbose(verbosity=true)
...
           collection.push(param_class.new(param_hash))
         else
```
er0ck, line 145: seems like this method is now doing two very distinct things.  should be broken up

lib/rototiller/utilities/command.rb
```diff
@@ -5,22 +5,62 @@ class Command
...
+
+      # determine the value of the command
+      if override_env && argument_override_env
+        "#{override_env.value} #{argument_override_env.value}"
+      elsif override_env
+        override_env.value
+      elsif argument_override_env
+        "#{name} #{argument_override_env.value}"
+      elsif argument
+        "#{name} #{argument}"
```
er0ck: does this imply that flags will be after args?  this probably won't work for BSDs (osx)

er0ck:
this needs more unit tests for the various combinations of args with and without envs that may or may not be set.
we should also cover the with and without flags cases because this will break some commands on some platforms (BSD).  maybe try the `ls` command with a mock for system return of `ls`.  or some other command that takes args and flags

zreichert:
@er0ck Updated

acceptance/lib/test_utilities.rb
```diff
@@ -0,0 +1,28 @@
...
+    (0...10).map { ('a'..'z').to_a[rand(26)] }.join.upcase
+  end
+
+  def set_random_env_on(host)
+    name = unique_env_on(host)
+    host.add_env_var(name, random_string)
+    return name
+  end
+
+  def unique_env_on(host)
```
er0ck, line 13: comment this please.  took me a few minutes to figure out what this was doing

lib/rototiller/utilities/command.rb
```diff
@@ -5,22 +5,37 @@ class Command
...
+
     # Creates a new instance of CommandFlag, holds information about desired state of a command
     # @param [Hash] attribute_hash hashes of information about the command
     # @option attribute_hash [String] :command The command
     # @option attribute_hash [String] :override_env The environment variable that can override this command
-    def initialize(attribute_hash)
-      # translate the keys from 'Command' to Env
-      if attribute_hash[:override_env]
-        @override_env = EnvVar.new({:name => attribute_hash [:override_env], :default => attribute_hash[:name]})
+    def initialize(h = {})
```
er0ck, line 23: why `h` instead of `attribute_hash`?
EnvVar still uses `attribute_hash`

PR #40, opened 2016-05-05
Add ability to ignore a command flag if default is nil

samwoods1:
I've got the optional param: 
  t.add_flag do |flag|
    flag.name = '--load-path'
    flag.default = nil
    flag.message = 'Additional load path for Beaker'
    flag.override_env = 'BEAKER_LOAD_PATH'
  end

And I wanted to be able to have it not included at all if the default was nil.  Maybe there's a better way to indicate that it is optional?  Would we prefer an additional 'optional' param for flag?

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,19 @@ Rototiller provides a Rake DSL addition called 'rototiller_task' which is a full
       end
     end
 
+    desc "do not include flag if the final value (either the default or override_env) is nil or empty and not required"
+    rototiller_task :test_flag_env do |task|
+      task.add_command do |cmd|
+        cmd.name = 'test'
+      end
+      task.add_flag do |flag|
+        flag.name         = '-f'
+        flag.default      = ''
+        flag.override_env = 'FLAG_VALUE'
+        flag.required     = false
+      end
+    end
+
     desc "override command argument values with environment variables"
     rototiller_task :test_arg_env do |task|
       task.add_command do |cmd|
--- a/acceptance/tests/rototillerTask/override_command_flag_with_env_var.rb
+++ b/acceptance/tests/rototillerTask/override_command_flag_with_env_var.rb
@@ -1,9 +1,11 @@
 require 'beaker/hosts'
 require 'rakefile_tools'
+require 'test_utilities'
 
 test_name 'C97798: existing workflows shall be supported for using ENV vars to override command flags' do
   extend Beaker::Hosts
   extend RakefileTools
+  extend TestUtilities
 
   def create_rakefile_task_segment(flags)
     segment = ''
@@ -95,12 +97,46 @@ end
 
       on(sut, "rake #{@task_name}", :accept_all_exit_codes => true) do |result|
         # exit code & no error in output
-        assert(result.exit_code == 1, 'The expected exit code 0 was not observed')
+        assert(result.exit_code == 1, 'The expected exit code 1 was not observed')
         assert_no_match(/error/i, result.output, 'An unexpected error was observed')
 
         command_flags.each do |flag|
           regex = /The CLI flag #{flag[:name]} needs a value.\nYou can specify this value with the environment variable #{flag[:override_env]}/
           assert_match(regex, result.stdout, "The expected output from rototiller was not observed")
+        end
+      end
+    end
+  end
+
+  step 'Test flags that are not required' do
+    override = random_string
+    command_flags = [
+        {:name => '--not-required', :override_env => 'required_override', :required => false},
+        {:name => '--not-required-with-default', :override_env => override, :default => 'def_val', :required => false},
+    ]
+
+    @task_name    = 'command_flags_that_stop_rake'
+    rakefile_contents = <<-EOS
+$LOAD_PATH.unshift('/root/rototiller/lib')
+require 'rototiller'
+
+Rototiller::Task::RototillerTask.define_task :#{@task_name} do |t|
+    #{create_rakefile_task_segment(command_flags)}
+    t.add_command({:name => 'echo'})
+end
+    EOS
+    rakefile_path = create_rakefile_on(sut, rakefile_contents)
+
+    step 'Execute task defined in rake task' do
+
+      on(sut, "rake #{@task_name} #{override}=''", :accept_all_exit_codes => true) do |result|
+        # exit code & no error in output
+        assert(result.exit_code == 0, 'The expected exit code 0 was not observed')
+        assert_no_match(/error/i, result.output, 'An unexpected error was observed')
+
+        command_flags.each do |flag|
+          regex = /The CLI flag #{flag[:name]} has no value assigned and will not be included./
+          assert_match(regex, result.stdout, "The expected output from rototiller for an optional flag was not observed.")
         end
       end
     end
--- a/lib/rototiller/task/rototiller_task.rb
+++ b/lib/rototiller/task/rototiller_task.rb
@@ -46,12 +46,13 @@ module Rototiller
       # @option args [String] :name The environment variable
       # @option args [String] :default The default value for the environment variable
       # @option args [String] :message A message describing the use of this variable
+      # @option args [Boolean] :required Is used internally by CommandFlag, ignored for a standalone EnvVar
       #
       # for block {|a| ... }
       # @yield [a] Optional block syntax allows you to specify information about the environment variable, available methods track hash keys
       def add_env(*args,&block)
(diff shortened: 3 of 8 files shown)


samwoods1:
Ok... So after working through this I kept running into various scenarios that didn't quite work, so I came up with some requirements:
- flag not required, default to 'not include', overridable by env_var
- flag not required, default value, overridable by env_var to different value
- flag not required, default value, overridable by env_var to 'not include'
- flag required, no default value, overridable by env_var
- flag required, default value, overridable by env_var

I think this approach fulfills all of the requirements:
- Add a new 'required' argument to add_flag that defaults to true. If set to true, everything behaves as it does currently.  If set to false then this would happen:
  - if the default is set to nil or empty string, or not set to anything then the flag will not be included in the command string unless overridden by an env_var
  - if the default is set to some value, but the env_var is set to empty string then the flag will not be included in the command string.

I'm working on updating the PR now.  Fixing unit tests.

er0ck:
> flag not required, default to 'not include', overridable by env_var

are you saying that you'd set {{:default}} to 'not include' and that string should somehow be 'magic' and make the flag disappear from the command string?

edit: oh, i think you are saying it's just by default not included in the command string.  but if you have an env_var for its value, it'd show up, followed by its value.
yeah.  i'm toying with it now.  i'll be sure to include these requirements.  thanks!

edit2: 
i think your requirements are good, but the API feels very complicated to me.  so complicated that i can't remember how it works, and i just read it.  we might just have to be verbose here, and make the API consistent, and lengthy, but discoverable and predictable.
e.g.: i'm toying with {{add_flag({:name, :default, :override_env, :boolean, :key_override_env}}
where we just have params that dictate if it's boolean and any env variable(s) that override the key/body of the flag.  i HATE my current naming of those new params, tho

er0ck:
added some labels just for kicks.  feel free to ignore yard docs and human docs, if it's too heavy, in this timeframe.
"Needs Human Docs" is for "human readable", markdown docs

er0ck:
for future reference... it might be advantageous to have the typos, etc in a separate (maint) commit, so if we have to back-out your other changes, we get to keep the speling fixes

er0ck:
👍 LGTM.  note, we are cutting some corners here, under the assumption that we need to ship this thing, and we're going to gut the API next release

lib/rototiller/task/rototiller_task.rb
```diff
@@ -46,12 +46,14 @@ def self.define_task(*args, &task_block)
       # @option args [String] :name The environment variable
       # @option args [String] :default The default value for the environment variable
       # @option args [String] :message A message describing the use of this variable
+      # @option args [Boolean] :required Indicates whether an error should be raised
+      # if the value is nil or empty string, vs not including the flag.
```
zreichert: Should this be 'vs not manage the environment variable'?

samwoods1: No... It really only makes sense in the context of a CommandFlag, but I need it in EnvVar since it is used internally by CommandFlag and contains the logic to determine what the final value for the flag will be.  Since we would never expect a user to specify :required for just an EnvVar, and only expect CommandFlag to use it internally when creating an EnvVar would we want to remove this from the documentation?

samwoods1:
@er0ck Just curious, what corners are we cutting?  If there's additional stuff required, let me know.

er0ck:
acceptance test would be great.  but i hesitate as the api will change so much, hopefully quickly.  then again, an acceptance test that tests the high level feature of being able to turn off an option and its value would be valuable now and later (v0.2.z)

i think i was mostly just trying (and failing) to state for posterity that this isn't really the best solution for the API but we've painted ourselves into a corner.  the RototillerTask api needs re-work before we move on to beaker tasks, etc.

samwoods1:
Added acceptance tests, which uncovered some stuff, so also fixed some logic and now convert 'true' string to boolean (acceptance tests were assuming everything was a string)

README.md
```diff
@@ -48,6 +48,19 @@ Rototiller provides a Rake DSL addition called 'rototiller_task' which is a full
       end
     end
 
+    desc "do not include flag if value is nil or empty and not required"
```
er0ck: nit: the word "value" is kinda ambiguous here

acceptance/tests/rototillerTask/override_command_flag_with_env_var.rb
```diff
@@ -105,4 +105,38 @@ def remove_reserved_keys(h)
       end
     end
   end
+
+  step 'Test flags that are not required' do
+    override = (0...10).map { ('a'..'z').to_a[rand(26)] }.join
```
er0ck: we have {{random_string()}} in the acceptance lib
but it upcases.  that upcase should probably be removed.  one can upcase their own lazy self.
anyway, use that random_string helper here.

will require quick update to {{command_arguments_with_override.rb}}  (add .upcase to random_string)

acceptance/tests/rototillerTask/override_command_flag_with_env_var.rb
```diff
@@ -105,4 +105,38 @@ def remove_reserved_keys(h)
...
+    step 'Execute task defined in rake task' do
+
+      on(sut, "rake #{@task_name} #{override}=''", :accept_all_exit_codes => true) do |result|
+        # exit code & no error in output
+        assert(result.exit_code == 0, 'The expected exit code 0 was not observed')
+        assert_no_match(/error/i, result.output, 'An unexpected error was observed')
+
+        command_flags.each do |flag|
+          regex = /The CLI flag #{flag[:name]} has no value assigned and will not be included./
+          assert_match(regex, result.stdout, "The expected output from rototiller was not observed")
```
er0ck: differentiate this assertion log message from the one above

lib/rototiller/utilities/env_var.rb
```diff
@@ -67,6 +73,11 @@ def message=(message)
     reset
   end
 
+  def required=(required)
+    @required = required
+    reset
+  end
```
er0ck, line 90: please comment wtf this is doing 😟

lib/rototiller/utilities/env_var.rb
```diff
@@ -1,10 +1,10 @@
 require 'rototiller/utilities/color_text'
 
 class EnvVar
-  MESSAGE_TYPES = {:nodefault_noexist=>0, :exist=>1, :default_noexist=>2}
+  MESSAGE_TYPES = {:nodefault_noexist=>0, :exist=>1, :default_noexist=>2, :not_required=>3}
   include ColorText
 
-  attr_accessor :var, :message, :default
+  attr_accessor :var, :message, :default, :required
```
er0ck: :nit these need yardoc comments, but they weren't there before, so i'm willing to kick it on down 🎢

README.md
samwoods1: Yeah, I was trying to say that it is the final value after the default and override_env are evaluated.  What would be a more self explanatory word for that?  final value?

lib/rototiller/utilities/env_var.rb
samwoods1, line 90: Added a comment above this block of variable assignment methods that are all doing the same thing.

er0ck, line 90: 🆗

acceptance/tests/rototillerTask/override_command_flag_with_env_var.rb
```diff
@@ -105,4 +107,38 @@ def remove_reserved_keys(h)
       end
     end
   end
+
+  step 'Test flags that are not required' do
+    override = random_string
+    command_flags = [
+        {:name => '--not-required', :override_env => 'required_override', :required => false},
+        {:name => '--not-required-with-default', :override_env => override, :default => 'def_val', :required => false},
```
zreichert, line 115: Sorry if I am being dense but in what circumstance does the :default => 'def_val' get used?  Followup question if I set override in the ENV does it use that value?

samwoods1, line 115: def_val would be used, except that I am setting the override to an empty string (line 132).
Yes, if you set override to something other than an empty string, then it will use that value instead.

zreichert, line 115: @samwoods1 Thanks for the clarification, makes sense.

zreichert, line 115: 👍  LGTM

PR #42, opened 2016-05-11
(QA-2445) Add support for setting boolean flags with rototiller

er0ck:
We currently do not have a way to turn on and off switches, aka:
boolean-flags/options. This change adds a switch parameter to add_flag
to specify is_boolean.  The override_env now operates on the option
name, and the default value can be used to default the switch to 'off'.

This is bad and we should feel bad. the API if very warty and needs
refactoring. (post v0.1.0)

### DIFF
--- a/acceptance/tests/rototillerTask/command_switches.rb
+++ b/acceptance/tests/rototillerTask/command_switches.rb
@@ -1,0 +1,94 @@
+require 'beaker/hosts'
+require 'rakefile_tools'
+
+test_name 'C97821: can set switches (boolean options) for commands in a RototillerTask' do
+  extend Beaker::Hosts
+  extend RakefileTools
+
+  test_filename = File.basename(__FILE__, '.*')
+
+  def create_rakefile_task_segment(flags)
+    segment = ''
+    flags.each.with_index do |flag,index|
+      sut.add_env_var(flag[:override_env], flag[:env_value]) if flag[:override_env]
+      if flag[:block_syntax]
+        segment += "t.add_flag do |flag|\n"
+        remove_reserved_keys(flag).each do |k, v|
+          segment += "  flag.#{k.to_s} = '#{v}'\n"
+        end
+        segment += "end\n"
+      else
+        segment += "  t.add_flag({"
+        remove_reserved_keys(flag).each do |k, v|
+          segment += ":#{k} => '#{v}',"
+        end
+        segment += "})\n"
+      end
+    end
+    return segment
+  end
+
+  def remove_reserved_keys(h)
+    hash = h.dup
+    [:block_syntax, :env_value].each do |key|
+      hash.delete(key)
+    end
+    return hash
+  end
+
+  command_flags = [
+      {:name => '--name1',                  :is_boolean => true},
+      {:name => '--name2',  :default => '', :is_boolean => true},
+      {:name => '--name3',                  :is_boolean => true, :override_env => 'NODEFAULT1',  :env_value => 'VAL1'},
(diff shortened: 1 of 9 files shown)


acceptance/tests/rototillerTask/command_switches.rb
```diff
@@ -0,0 +1,94 @@
...
+\e[33mThe CLI switch '--name5' will NOT be used.\e[0m
+\e[33mThe CLI switch '--name6' will NOT be used.\e[0m
+\e[32mThe CLI switch '--name7' will be used.\e[0m
+\e[33mThe CLI switch '--name8' will NOT be used.\e[0m
+\e[32mThe CLI switch 'VAL3' will be used.\e[0m
+\e[32mThe CLI switch 'VAL4' will be used.\e[0m
+\e[33mThe CLI switch '--nameB' will NOT be used.\e[0m
+\e[33mThe CLI switch '--nameC' will NOT be used.\e[0m
+
+--name1 VAL1 VAL2 --name7 VAL3 VAL4
```
zreichert, line 88: Should VAL1 VAL2 VAL3 VAL4 be used in the command?  It looks like all of these are set to :is_boolean => true.

acceptance/tests/rototillerTask/command_switches.rb
```diff
@@ -0,0 +1,94 @@
...
+  rakefile_path = create_rakefile_on(sut, rakefile_contents)
+
+  step 'Execute task defined in rake task' do
+    on(sut, "rake #{@task_name}", :accept_all_exit_codes => true) do |result|
+      # exit code & no error in output
+      assert(result.exit_code == 0, 'The expected exit code 0 was not observed')
+      assert_no_match(/error/i, result.output, 'An unexpected error was observed')
+
+      # i plainly refuse to re-implement rototiller's is_boolean option logic here
+      expected_out = <<-HERE
```
zreichert, line 74: nit: pick EOS or HERE

acceptance/tests/rototillerTask/command_switches.rb
er0ck, line 88: that's why VAL1, etc got replaced into the switch itself.  so name3 and name4 switch bodies got replaced with VAL1 and VAL2, etc.  naming is hard.

yeah, all these are boolean.  this is the switch test

spec/rototiller/utilities/command_flag_spec.rb
```diff
@@ -105,9 +122,116 @@
...
+      context 'should set switch to env_override when env is set with a :default' do
+        let(:args) { {:name => flag, :is_boolean => true, :default => value, :override_env => override_env} }
+        let(:env_val) { "env_#{value}" }
+
+        it 'should report the switch name' do
+          ENV[override_env] = env_val
+          expect(subject.flag).to eq(env_val)
+        end
+
+        # FIXME: repeated ENV setting here is problematic
```
zreichert, line 222: I don't understand what is problematic about this.  Did you want to set the ENV in a before?

er0ck, line 222: problematic isn't really the right word. 
it's repetitive and error prone.  
eg, when more tests are added, people might set the env wrong, etc.  

i just couldn't get it to work in before, nor let

zreichert, line 222: Perhaps we should build a helper that more actively manages the ENV.

acceptance/tests/rototillerTask/command_switches.rb
zreichert, line 88: I think we should ask for input on this functionality from users of the beta.

PR #43, opened 2016-05-12
(maint) use correct library when building gem

er0ck:
Before this change, acceptance would use any installed rototiller gem
for its version string when building the gem for acceptance. This would
cause issues if not using bundler, or had some old version of the gem.
And we have this info right relatively here, let's use it.


samwoods1:
👍

er0ck:
@zreichert 👀 please?

PR #48, opened 2016-06-17
(QA-2521) define all permutations for v2 API

er0ck:
[skip ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,140 @@ The known changes include (not comprehensive):
 * adding some sort of env_var type so one does not have to use the `:required` parameter for `#add_flag`
 * the above will allow for multiple commands in a task with independent option, switch, and environment variable tracking
 
-Examples (see the [Use](#use) section for outputs):
+Rototiller has 4 main _types_ of arguments that can be passed to a command in a task. `RototillerTasks` can accept multiple commands.  Each of these argument types has a similar API that looks similar to `add_command()`.
+### All permutations of v2 API (remove and refactor into useful doc sections below upon testing, merge-up to stable)
+
+* all things that can take multiples should use add\_ as precursor to method name
+* all things that only take one should use set\_ as precursor to method name?
+    require 'rototiller'
+
+    ## all task methods
+    rototiller_task :name do |t|
+      t.add_command # t.add_cmd? me no likey
+      t.add_env
+    end
+    rototiller_task do |t|
+      t.set_name = 'string_name' # should this be validated??  e.g.: spaces, etc
+      t.add_command
+      t.add_env
+    end
+
+
+    ## all task's add_env invocations with just name
+    t.add_env('env_name') #required, default messaging
+    t.add_env({:name => 'env_name'})
+    t.add_env :env_name
+    t.add_env 'env_name' # implicitly allowed by ruby
+    t.add_env do |e|
+      e.name
+    end
+
+    ## all task's add_env invocations with name, message
+    #t.add_env('env_name')  # impossible
+    t.add_env({:name => 'env_name', :message})
+    t.add_env :env_name do |e|
+    t.add_env 'env_name' do |e|  # should we do this too?
+      e.set_message
+    end
+    t.add_env do |e|
+      e.name
+      e.message
+    end
+
+    ## all task's add_env invocations with name, value
+    #t.add_env('env_name')  # impossible
+    t.add_env({:name => 'env_name', :default/:value =>}) # optional; default messaging
+    t.add_env :env_name do |e|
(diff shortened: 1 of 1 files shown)


er0ck:
notes from 6/21 meeting:
maybe set_message() goes away (it never existed?) and we can only user-set messaging on add_env()?

otherwise we have messaging inception

README.md
```diff
@@ -31,7 +31,140 @@ The known changes include (not comprehensive):
...
+    end
+
+
+    ## all task's add_env invocations with just name
+    t.add_env('env_name') #required, default messaging
+    t.add_env({:name => 'env_name'})
+    t.add_env :env_name
+    t.add_env 'env_name' # implicitly allowed by ruby
+    t.add_env do |e|
+      e.name
```
er0ck, line 59: set_name()?

add a note about which takes precedent, apply same precedent to tasks, options, args, etc

README.md
```diff
@@ -31,7 +31,140 @@ The known changes include (not comprehensive):
...
+
+    ## all task's add_env invocations with name, message
+    #t.add_env('env_name')  # impossible
+    t.add_env({:name => 'env_name', :message})
+    t.add_env :env_name do |e|
+    t.add_env 'env_name' do |e|  # should we do this too?
+      e.set_message
+    end
+    t.add_env do |e|
+      e.name
```
er0ck, line 70: set_name()?

README.md
```diff
@@ -31,7 +31,140 @@ The known changes include (not comprehensive):
...
+      e.message
+    end
+
+    ## all task's add_env invocations with name, value
+    #t.add_env('env_name')  # impossible
+    t.add_env({:name => 'env_name', :default/:value =>}) # optional; default messaging
+    t.add_env :env_name do |e|
+    t.add_env 'env_name' do |e|  # should we do this too?
```
er0ck, line 78: no dashes in symbols.  create some conversion inside?  iw, beaker does this with peril

README.md
```diff
@@ -31,7 +31,140 @@ The known changes include (not comprehensive):
...
+      e.default/value  # does value imply the env will be set by rototiller?  does default NOT?
+    end
+    t.add_env do |e|
+      e.name
```
er0ck, line 82: set_name()

README.md
```diff
@@ -31,7 +31,140 @@ The known changes include (not comprehensive):
...
+    t.add_command do |c|
+      c.name = 'echo'
+    end
+
+    ## all task's add_command invocations with name (string), message
+    #t.add_command('echo --blah my name is ray', 'message') # ArgumentError
+    t.add_command({:name => 'echo', :message => 'blah'})
+    t.add_command :echo
+    t.add_command 'echo' do |c|
+      c.name = 'echo' # # nomethod error?
```
er0ck, line 115: same precedent as above. should we support set_name inside the blocks?

README.md
```diff
@@ -31,7 +31,140 @@ The known changes include (not comprehensive):
...
+    #t.add_command 'echo' do |c|
+    #  c.message = 'blah'
+    #end
+    t.add_command do |c|
+      c.name 'echo' do |n|
+        n.add_env
+      end
+      c.add_arg 'some_arg' do |a|
+        a.add_env
+        a.message
```
er0ck, line 136: set_message

README.md
```diff
@@ -31,7 +31,140 @@ The known changes include (not comprehensive):
...
+          e.set_value
+        end
+        o.message
+      end
+      c.add_switch '--some_switch' do |s|
+        s.add_env 'env_name' do |e|
+          e.set_name
+          e.set_message
+          e.set_value
+        end
```
er0ck, line 158: s.add_arg() -> nomethod, arg error

README.md
```diff
@@ -31,7 +31,140 @@ The known changes include (not comprehensive):
...
+        s.message
+      end
+    end
+
+    #should we be able to add an env for any given message?  i don't see a use case, we should probably just save users from themselves here.
+
```
er0ck, line 164: message should be called "description" to match rake `desc`?  even if inside block vs outside

er0ck:
how to name methods and hash names for levels?
make them the same!

`set_singular`
and 
`add_multiple`

so add_blah takes arrays
set_blah takes non-arrays

er0ck:
if we get a complicated hash, with args, options, commands, issue warning stating that ordering won't be preserved

er0ck:
should we allow `desc` inside blocks above an `add_` or `set_` method, so it's similar to how we create descriptions in rake?

this depends partly on whether we think descriptions (in options, commands, etc) can also take `add_env` (as there would be no way to add_env to a description set outside a block, but maybe that's just a limitation in using `desc`)

PR #49, opened 2016-07-13
(qa-2478) refactor rototiller acceptance repeated code

er0ck:
- abstract the rakefile segment generation
  - The rakefile segment generation method was very similar across tests. Move it to a library. It could probably use some clarity and more fanciness, but this works for now and should allow us to more easily create tests for api v2.
- create unique Rakefile per test
  - Previously we were just putting a Rakefile in /root. This is great for using the default rake behavior of finding Rakefiles. but is difficult to debug when tests fail.  Now we place a unique-per-run Rakefile in /tmp and specify that file for rake to run.
- abstract remove_reserved_keys out of tests
- abstract rake task execution out of tests
- rename some "flag" usage to "switch" in tests
- abstract rakefile headers, remove from tests
  - the header for the Rakefiles used in tests were repeated across all tests. Here we abstract this into a module
- remove step blocks in modules
  - steps with blocks in modules causes beaker to not un-indent the logging
    properly, remove these.  :-(

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -28,6 +28,12 @@ rototiller_task :acceptance => [:generate_host_config] do |t|
 
   # with new block syntax
   t.add_flag do |flag|
+    flag.name = '--log-level'
+    flag.default ="verbose"
+    flag.message = 'beaker log-level'
+    flag.override_env = 'BEAKER_LOG_LEVEL'
+  end
+  t.add_flag do |flag|
     flag.name = '--hosts'
     flag.default = 'acceptance/hosts.cfg'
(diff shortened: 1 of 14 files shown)


er0ck:
@seangriff you could help by reviewing this as well, if you have time.

er0ck:
@samwoods1 if you have time to review, that'd be cool

er0ck:
@seangriff, thoughts on this PR?

samwoods1:
Much goodness 👍

er0ck:
@zreichert can you push go on this?

PR #51, opened 2016-08-05
(QA-2549) add multiple command handling

er0ck:


lib/rototiller/task/rototiller_task.rb
```diff
@@ -49,29 +41,23 @@ def self.define_task(*args, &task_block)
...
-        attributes = [:name, :default, :message, :override_env, :required, :is_boolean]
-        add_param(@flags, CommandFlag, attributes, args, &block)
+      # @yield [a] Optional block syntax allows you to specify information about the environment variable, available methods match hash keys
+      def add_env(*args, &block)
+        raise ArgumentError.new("#{__method__} takes a block or a hash") if !args.empty? && block_given?
+        # this is kinda annoying we have to do this for all params? (not DRY)
+        #   have to do it this way so EnvVar doesn't become a collection
+        #   but if this gets moved to a mixin, it might be more tolerable
+        if block_given?
+          empty_arg = []
```
er0ck: send empty hash, dummy!

er0ck:
@samwoods1 can you take a look at this as well? it would probably be good if you understood the direction we are going with the API and the architecture here.

Rakefile
```diff
@@ -24,57 +24,88 @@ end
...
+    #flag.override_env = "BEAKER_PRE_SUITE"
+  #end
+  #t.add_flag do |flag|
+    #flag.name = '--tests'
+    #flag.default = 'acceptance/tests'
+    #flag.message = 'The path to the tests you want beaker to run'
+    #flag.override_env = 'BEAKER_TESTS'
+  #end
+
+  t.add_command({:name => 'beaker --log-level verbose --hosts acceptance/hosts.cfg --preserve-hosts --keyfile ~/.ssh/id_rsa-acceptance --load-path acceptance/lib/ --pre-suite acceptance/pre-suite --tests acceptance/tests/', :override_env => 'BEAKER_EXECUTABLE'})
```
zreichert, line 74: Is it not possible to use the new API at this point?

er0ck: that is using the new API.  just not in the best way.  add_flag is totally hosed right now, i forget why i had to do it that way, i think due to moving all the files around and adding command_collections.

lib/rototiller/task/collections/param_collection.rb
```diff
@@ -0,0 +1,82 @@
...
+
+        @collection.each do |param|
+
+          filtered.push(param) if filters.all? do |method, value|
+
+            if param.respond_to?(method)
+              if param.send(method).nil?
+                value.nil?
+              else
+                param.send(method).to_s =~ /#{value.to_s}/
```
zreichert, line 52: Do you want to do a match here? It seems like we should do an actual comparison.

``` ruby
param.send(method).to_s == value.to_s
```

Would there be a case where `param.send(method).to_s` would return a subset of what `value.to_s` returns?

I realize now that I think I wrote this originally 😨

lib/rototiller/task/params/command_flag.rb
```diff
@@ -0,0 +1,129 @@
+require 'rototiller/utilities/color_text'
+require 'rototiller/task/params/env_var'
+
+module Rototiller
+  module Task
+
+    class CommandFlag
```
zreichert, line 7: Why does `CommandFlag` not inherit from `RototillerParam`?

lib/rototiller/task/collections/param_collection.rb
er0ck: yeah, i didn't change this (portion), the file just moved.

lib/rototiller/task/params/command_flag.rb
er0ck: because it is not a part of the intent of this ticket.  future tickets are to move and fix add_option/switch

lib/rototiller/task/block_handling.rb
```diff
@@ -1,20 +1,28 @@
 module Rototiller
   module Task
+    module BlockHandling
 
-    class Block_syntax
+      class BlockSyntax
```
zreichert, line 5: YAY Ruby style!

lib/rototiller/task/params/command.rb
```diff
@@ -40,6 +47,24 @@ def initialize(attribute_hash = {})
           @argument = attribute_hash[:argument]
         end
       end
+
+      #FIXME: move the delete_if stuff to its own private method
```
zreichert: Agree

lib/rototiller/task/rototiller_task.rb
```diff
@@ -1,36 +1,29 @@
...
-require 'rototiller/task/block_handling'
+require 'rototiller/task/collections/command_collection'
 require 'rake/tasklib'
 
 module Rototiller
   module Task
+
     class RototillerTask < ::Rake::TaskLib
       #TODO rename instance vars and methods to not match sub blocks
+      #  huh?
```
zreichert: Is this talking about how names and methods don't line up across classes?

lib/rototiller/task/params/command.rb
er0ck: done, in subsequent commits :-D

lib/rototiller/task/collections/param_collection.rb
```diff
@@ -52,6 +59,12 @@ def filter_contents(filters={})
         filtered
       end
 
+      #@private
+      def allowed_class
+        EnvVar
```
zreichert: Should we include a #TODO noting that this is required for the tests and should be removed once fixed?

lib/rototiller/task/collections/param_collection.rb
```diff
@@ -60,11 +58,6 @@ def filter_contents(filters={})
       end
 
       #@private
-      def allowed_class
```
zreichert: Ok, see you did this now.

er0ck: oh hells yes.  moving this into the child collections was so satisfying.  :-D

lib/rototiller/task/collections/command_collection.rb
```diff
@@ -0,0 +1,14 @@
+require 'rototiller/task/collections/param_collection'
+require 'rototiller/task/params/command'
+
+module Rototiller
+  module Task
+
+    class CommandCollection < ParamCollection
+      def allowed_class
+        Command
+      end
```
er0ck, line 10: i actually thought i could just infer the class name here, but it was kludgy by finding `.class` and then removing `'Collection'`  :-(
that would have made it so we didn't need the children

lib/rototiller/task/rototiller_task.rb
```diff
@@ -1,36 +1,29 @@
...
       # Whether or not to fail Rake when an error occurs (typically when
       # examples fail). Defaults to `true`.
       attr_accessor :fail_on_error
-
       # A message to print to stderr when there are failures.
       attr_accessor :failure_message
 
       def initialize(*args, &task_block)
         @name          = args.shift
+        # TODO: make fail_on_error per command?
```
samwoods1: 👍 I think this should for sure be per command.

lib/rototiller/task/collections/command_collection.rb
zreichert: We could just do an `attr_reader` on PramCollection and assign the instance variable in an `initialize`.  We are probably splitting hairs here thought.

lib/rototiller/task/collections/param_collection.rb
er0ck: it's moved to the child classes in subsequent commits

lib/rototiller/task/rototiller_task.rb
er0ck: created QA-2621 for this

samwoods1:
👍

lib/rototiller/task/collections/command_collection.rb
er0ck: yeah, i also don't like that.  it means whatever uses CommandCollection has to know what class it allows.  it should know itself.
oh.  maybe you mean it's a reader, and the initializer in here sets it, then calls super?
yeah, seems like more hoops to jump through without gaining anything.  this is just an explicit reader.  
can you set a value when doing readers?
`attr_reader :allowed_class = Command`
nope:

```
SyntaxError: (irb):5: syntax error, unexpected '=', expecting keyword_end
attr_reader :other = 'value'
```

zreichert: I think you would do something like this.  

``` Ruby
class ParamCollection
  attr_reader :allowed_class

  def initialize 
    @allowed_class = FooBar
  end
end
```

``` Ruby
class CommandCollection < ParamCollection

  def initialize 
    @allowed_class = FooBaz
  end
end
```

er0ck: yeah.  that would work.  but i don't see how it's better, nor more clear.  and it's longer

zreichert: I disagree about its clarity.  

I was trying to address this comment you had originally made.  

> i actually thought i could just infer the class name here, but it was kludgy by finding .class and then removing 'Collection' :-(
> that would have made it so we didn't need the children

I agree that this would be kludgy and was just trying to provide an alternative.

README.md
```diff
@@ -32,6 +32,131 @@ The known changes include (not comprehensive):
...
+
+Finally navigate to http://0.0.0.0:8808/ to view the documentation
+
+## Maintainers
+* [Zach Reichert](zach.reichert@puppetlabs.com), github:[zreichert](https://github.com/zreichert), jira:zach.reichert
+* [Eric Thompson](erict@puppetlabs.com), github:[er0ck](https://github.com/er0ck), jira:erict
+* [QA](qa-team@puppetlabs.com)
+
+
+## abandon hope, all ye who enter here
```
zreichert, line 159: he he, yeah there be dragons here

Rakefile
```diff
@@ -24,57 +24,88 @@ end
...
+    puts output
+    if output =~ /\[warn\]|\[error\]/
+      begin # prevent pointless stack on purposeful fail
+        fail "Errors/Warnings during yard documentation generation"
+      rescue Exception => e
+        puts 'Yardoc generation failed: ' + e.message
+        exit 1
+      end
+    end
+    `yard graph --full | dot -Tpng -o #{DOCS_DIR}/rototiller_class_graph.png`
```
zreichert, line 109: dot is not installed by default (OSX 10.11.6)

``` bash
ztr:rototiller zachreichert$ bundle exec rake yard
rm -rf yard_docs
Files:          13
Modules:         7 (    7 undocumented)
Classes:        11 (   11 undocumented)
Constants:       2 (    2 undocumented)
Attributes:     18 (    0 undocumented)
Methods:        30 (    9 undocumented)
 57.35% documented
sh: dot: command not found 
```

Can't seem to find 'dot' with homebrew :(

er0ck: fixed, i believe.  added graphviz as dev dependency

zreichert: Unfortunately I don't think that graphviz adds the 'dot' command.  At least it didn't on my Mac.

PR #54, opened 2016-09-16
(maint) acceptance: add platform axis support

er0ck:
The current job templates we use in CI require LAYOUT as the variable for platform.
This will help us test on the platforms with varying ruby versions.  no need for a ruby axis, installing rvm on SUTs, etc
corresponding PR here: https://github.com/puppetlabs/ci-job-configs/pull/1749

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,8 +13,7 @@ end
 task :test => [:check_test]
 
 task :generate_host_config do |t, args|
-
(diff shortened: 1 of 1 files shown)


er0ck:
@zreichert @samwoods1 FYI.

zreichert:
@er0ck So is this just moving from `TEST_TARGET` to `LAYOUT` because of our template in JJB?

er0ck:
@zreichert yes

PR #55, opened 2016-09-19
(QA-2637) command: #add env

er0ck:
- add #add_env to command
- it will overwrite command name with its value
- FIXME: we mainly duplicate add_env here from rototiller_task, it can
  be put in a module
- add #to_str for EnvVar so it can be printed and find its value
- add #last to env_collection so one can easily find the last known
  value of a collection of EnvVars.
- block_handling needed empty methods defined for nested blocks and
  hashes. When a Command is created and has a nested add_env, for
  instance, the add_env must be both a setter and a method (getter), even
  though it is not actually called.
- ensure we return the EnvVar or Command when using #add_env or
  #add_command
- EnvVar (and other params) need their attribute names to match their
  allowed attributes because we now must yield using EnvVar objects, so
  the blocks populate the actual EnvVar
  ee96188

### DIFF
--- a/lib/rototiller/task/params/command.rb
+++ b/lib/rototiller/task/params/command.rb
@@ -6,6 +6,15 @@ require 'rototiller/task/block_handling'
 module Rototiller
   module Task
 
+    # The Command class to implement rototiller command handling
+    #   via a RototillerTask's #add_command
+    # @since v0.1.0
+    # @attr [String] name The name of the command to run
+    # @attr_reader [Struct] result A structured command result
+    #    contains members: output, exit_code and pid (from Open3.popen2e)
+    # @attr [String] argument Command argument (this will change to add_argument in the next release)
+    # @attr [String] argument_override_env The environment variable (if any) users can employ to override the above argument
+    #    (this will change to Argument's #add_env in the next release)
     class Command < RototillerParam
       include Rototiller::ColorText
       include BlockHandling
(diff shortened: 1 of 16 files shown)


lib/rototiller/task/params/command.rb
```diff
@@ -56,6 +61,37 @@ def initialize(args={}, &block)
...
+        else
+          #TODO: test this with array and non-array single hash
+          args.each do |arg| # we can accept an array of hashes, each of which defines a param
+            error_string = "#{__method__} takes an Array of Hashes. Received Array of: '#{arg.class}'"
+            raise ArgumentError.new(error_string) unless arg.is_a?(Hash)
+            new_env_var = EnvVar.new(arg)
+            @env_vars.push(new_env_var)
+          end
+        end
+        @name = @env_vars.last if @env_vars.last
```
zreichert, line 91: Will `@name` be an EnvVar?
I think on line 21 we say the method `name` will return a String.

lib/rototiller/task/params/command.rb
```diff
@@ -56,6 +61,37 @@ def initialize(args={}, &block)
...
+        return new_env_var
```
zreichert: What do we do with the returned `EnvVar` ?

lib/rototiller/task/params/env_var.rb
```diff
@@ -5,28 +5,29 @@
 module Rototiller
   module Task
 
+    # The main EnvVAr type to implement envrironment variable handling
```
zreichert: nit: typo EnvVar not EnvVAr

lib/rototiller/task/params/command.rb
er0ck: EnvCollection's #last always returns a string.

lib/rototiller/task/params/command.rb
er0ck: you are right!  we don't need the returns.  

this half-way fixed a different issue that the `yield self` now fixes all the way

er0ck: crap.  we do need them in certain cases ❓ 

```
  1) Rototiller::Task::RototillerTask#add_command #add_env can override command name with env_var
     Failure/Error: c.add_env({:name => 'BLAH'})

     NoMethodError:
       undefined method `add_env' for [{:name=>"echo"}]:Array
     # ./spec/rototiller/task/rototiller_task_spec.rb:147:in `block (5 levels) in <module:Task>'
```

in this case the previous add_env is returning an Array of the hash (because it's just floating there as the last statement

er0ck: yeah, we need the returns in rototiller_task because of the difference in the way we handle collections in that context.  inside of params the returns should not be required.

lib/rototiller/task/params/command.rb
er0ck: maybe implicitly returning a string is a bit confusing and we should move the `#to_s` to here?  if so i'd like to do that later...

PR #66, opened 2016-11-11
(QA-2690) stop on any nested env var

zreichert:


er0ck (changes requested):
amend commits that contain (WIP) or no header.  edit: still.  fixup your commit msgs, please

Gemfile
```diff
@@ -12,6 +12,7 @@ group :system_tests do
   #gem 'beaker', :path => "../../beaker/"
   gem 'beaker'               ,'~> 2.22'
   gem 'beaker-hostgenerator'
+  gem 'public_suffix', '<= 1.4.6'
```
er0ck, line 15: alignment is off with above version string

acceptance/lib/test_utilities.rb
```diff
@@ -31,8 +31,8 @@ def execute_task_on(host, task_name=nil, rakefile_path=nil)
     command = "bundle exec rake #{task_name}"
     command = command + " --rakefile #{rakefile_path}" if rakefile_path
     on(host, command, :accept_all_exit_codes => true) do |result|
-      assert(result.exit_code == 0, "Unexpected exit code: #{result.exit_code}")
-      assert_no_match(/error/i, result.output, "An unexpected error was observed: '#{result.output}'")
+      #assert(result.exit_code == 0, "Unexpected exit code: #{result.exit_code}")
+      #assert_no_match(/error/i, result.output, "An unexpected error was observed: '#{result.output}'")
```
er0ck: please make this optional, hopefully default to true, if we think the negative tests will use this less frequently

acceptance/tests/rototillerTask/command_arguments_with_override.rb
```diff
@@ -27,6 +29,7 @@
     task_name    = test_filename + "#{tasks.length}"
     override_env = test_filename.upcase + random_string
 
+    #this task has no default for the ENV I think this should stop
```
er0ck: why wouldn't it take the name of the parent (name of the argument) as the env default?

acceptance/tests/rototillerTask/command_arguments_with_override.rb
```diff
@@ -136,6 +139,7 @@
   step 'create Rakefile, execute tasks' do
     rakefile_path = create_rakefile_on(sut, rakefile_contents)
 
+    require 'pry';binding.pry
```
er0ck: remove me

acceptance/tests/rototillerTask/env_var/negative/attempt_to_defign_an_env_with_invalid_characters.rb
```diff
@@ -0,0 +1,43 @@
+require 'beaker/hosts'
```
er0ck: speled `define` rong in the filename

acceptance/tests/rototillerTask/env_var/stopping_envs/multiple_stopping_envs_are_reported.rb
```diff
@@ -0,0 +1,64 @@
...
+      assert_no_match(/RUNNING/, result.stdout, "The command ran when it wasn't expected to")
+
+      #TODO what should this be????
+      rototiller_output_regex = //
+      assert_msg = 'The expected output was not observed'
+      assert_match(rototiller_output_regex, result.stdout, assert_msg)
+      assert(result.exit_code == 1, 'The expected error message was not observed')
+    end
+  end
+end
```
er0ck, line 62: i think all these tests should be combined into a single env_var stopping test.  could be single task for the envs that don't stop it.  and might have be multiple tasks for each one that will stop.  but should all be in same rakefile and just run the various tasks.  
see: https://github.com/puppetlabs/rototiller/blob/master/acceptance/tests/rototillerTask/command_arguments_with_override.rb

lib/rototiller.rb
```diff
@@ -1,2 +1,3 @@
+$LOAD_PATH.unshift('/Users/zachreichert/puppet_repos/rototiller/lib')
```
er0ck: yeah, this prolly won't work :-D

lib/rototiller/task/params/command.rb
```diff
@@ -160,6 +159,11 @@ def run
         @result
       end
 
+      def stop
```
er0ck, line 165: should this be `#stop?`

lib/rototiller/task/params/env_var.rb
```diff
@@ -25,11 +25,12 @@ class EnvVar < RototillerParam
       attr_accessor :required
       # FIXME: does (api) user need to read this directly?
       attr_reader :message_level
-      attr_reader :stop
       attr_reader :value
       attr_accessor :message
       attr_accessor :set_env
 
+      attr_reader :stop
```
er0ck, line 31: any reason this moved and is not with the other readers?

lib/rototiller/task/params/env_var.rb
```diff
@@ -153,14 +133,12 @@ def set_message_level
...
-          @stop = true
         when MESSAGE_TYPES[:exist], MESSAGE_TYPES[:default_noexist], MESSAGE_TYPES[:not_required]
           @message_level = :info
         else
           raise 'EnvVar: message type not supported'
         end
       end
     end
-
   end
```
er0ck, line 145: nit: i like this space to know where the module and class end.

lib/rototiller/task/params/option.rb
```diff
@@ -40,6 +40,11 @@ def add_argument(*args, &block)
       def to_str
         [@name.to_s, @arguments.to_s].compact.join(' ')
       end
+
+      def stop
+        return true if @arguments.stop?
+        return true unless @name
+      end
```
er0ck, line 50: wouldn't this also stop if it has an env that needs it to stop?

seems like this should be like `return true if [@arguments, @env_vars].any?{ |collection| collection.stop? }`

er0ck: i guess just because we want to stop if they don't receive a name? 
this works (doesn't stop) with `@name = ''`, yeah?

lib/rototiller/task/rototiller_task.rb
```diff
@@ -91,17 +91,20 @@ def add_command(*args, &block)
...
+
+      def stop_task?
         exit_code = 1
         exit exit_code if @env_vars.stop? || @commands.stop?
       end
 
       # @private
       def run_task
         print_messages
+        stop_task?
```
er0ck, line 108: nice

lib/rototiller/task/params/command.rb
```diff
@@ -160,6 +159,11 @@ def run
...
+        return true if [@switches, @options, @arguments].any?{ |collection| collection.stop? }
```
er0ck, line 166: shouldn't this check `env_vars` also?

acceptance/tests/rototillerTask/env_var/negative/attempt_to_defign_an_env_with_invalid_characters.rb
er0ck: i think there is another test with a speling mistake in the name as well.

lib/rototiller/task/params/command.rb
er0ck: needs yard-docs

lib/rototiller/task/params/option.rb
er0ck: needs yard-docs

lib/rototiller/task/params/switch.rb
```diff
@@ -52,6 +52,10 @@ def add_env(*args, &block)
         @name = @env_vars.last if @env_vars.last
       end
 
+      def stop
+        true unless @name
+      end
+
```
er0ck, line 61: needs yard-docs

lib/rototiller/task/rototiller_task.rb
```diff
@@ -91,17 +91,20 @@ def add_command(*args, &block)
       private
 
       # @private
-      # FIXME: separate out the exit stuff
       def print_messages
         puts @commands.format_messages
         puts @env_vars.format_messages
+      end
+
+      def stop_task?
```
er0ck, line 100: please put `# @private` above this

spec/rototiller/task/rototiller_task_spec.rb
```diff
@@ -158,6 +158,7 @@ def described_verbose(verbose)
               # this won't yet be set before add_command completes. is this okay?
               #expect(c.name).to eq('my_shiny_new_command')
             end
+            #not sure why this is calling exit twice
```
er0ck: because `described_run_task` causes the task to run

er0ck:
I added the human docs label.  I think this needs some documentation near add_env to describe when and why the task will stop, e.g.: when is an env `required`?

lib/rototiller.rb
zreichert: oops used for local debugging

lib/rototiller/task/params/command.rb
zreichert: Currently collections implement `stop?` and params implement `stop`.  We could change this but this is how it worked before.

acceptance/tests/rototillerTask/env_var/stopping_envs/multiple_stopping_envs_are_reported.rb
zreichert: I disagree, these may be verbose but they are easy to debug.  These tests are not expensive in terms of time and separating them out gives us a fine grained view into potential failures we may encounter.

acceptance/tests/rototillerTask/command_arguments_with_override.rb
zreichert: It would, I wrote this before we decided on this new behavior.  I'll remove this comment.

lib/rototiller/task/params/command.rb
er0ck: wrong question?  i believe this has to stop if any env_var says to stop
so... `|| @env_vars.any? { |collection| collection.stop }`

unless you change it to stop? as in my suggestion above and maybe you answer here

lib/rototiller/task/params/env_var.rb
```diff
@@ -127,22 +107,22 @@ def reset
       end
 
       # @private
-      def check
+      def env_value_provided_by_user?
         # its possible that name could not be set
-        ENV.key?(@name) if @name
+        true if (ENV.key?(@name) if @name)
```
er0ck: what happens if ENV is not set and `!@name`?
maybe this always has a name, but i'd like to see this changed to `return true if...` followed by `return false`

acceptance/lib/test_utilities.rb
zreichert: I can make this change but I think it will apply to a majority of the tests.  All tasks that stop exit 1 and message the word 'error'

lib/rototiller/task/params/option.rb
zreichert: An env_var can thing the task needs to stop, however if a param has a name it no longer stops.  The only time we ask if a collection of env_vars should stop is when it is attached to task.

lib/rototiller/task/params/command.rb
zreichert: See other answer. RE: stopping env collections

lib/rototiller/task/params/env_var.rb
zreichert: An env var must have a name.  We raise an error if one is not provided.  
`raise(ArgumentError, 'A name must be supplied to add_env') unless @name`

lib/rototiller/task/params/env_var.rb
zreichert: No, just an oversight.  Moved it up.

acceptance/lib/test_utilities.rb
er0ck: ok.  so no default, express what should happen in the call?

acceptance/tests/rototillerTask/env_var/stopping_envs/multiple_stopping_envs_are_reported.rb
er0ck: the tests def. need to be simplified then.  remove options if we aren't using options, etc.  the tests are difficult to read and know what we're testing and how

lib/rototiller/task/params/option.rb
er0ck: ah.  so in commands it's really the params that stop.  because if they have a default and there is no env_var it continues.  got it.

lib/rototiller/task/params/env_var.rb
er0ck: ok.  largely a nit.  just found it less than readable and i had to go check what the other code did to make sense of this

zreichert: Legit

spec/rototiller/task/rototiller_task_spec.rb
```diff
@@ -1,4 +1,4 @@
-require 'spec_helper'
+ require 'spec_helper'
```
er0ck: spurious space

acceptance/tests/rototillerTask/env_var/envs_that_do_not_stop/default_prefered_over_name_of_parent.rb
```diff
@@ -0,0 +1,50 @@
...
+  @block_syntax = 'block_syntax'
+
+  block_body = {
+      :add_command => {
+          :name => 'echo OOPS_THIS_IS_BROKEN',
+          :add_env => env_with_default,
+          :message => 'I am a message for the command',
+          :add_switch => {:name => '--switch'},
+          :add_option => { :name => '--option',
+                           :add_argument => {:name => 'option_argument'},
```
er0ck: can we remove these 4 lines?  they make me wonder wth this test is doing

acceptance/tests/rototillerTask/env_var/envs_that_do_not_stop/default_prefered_over_name_of_parent.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  rakefile_contents = <<-EOS
+#{rototiller_rakefile_header}
+Rototiller::Task::RototillerTask.define_task :#{@block_syntax} do |x|
+  #{RototillerBodyBuilder.new(block_body)}
+end
+  EOS
+  rakefile_path = create_rakefile_on(sut, rakefile_contents)
+
+  #add env to command
```
er0ck: what's this comment for?  it's confusing

spec/rototiller/task/rototiller_task_spec.rb
```diff
@@ -245,6 +245,23 @@ def described_verbose(verbose)
...
+          task.add_command(command)
+          expect{ described_run_task }.to output(/#{validation}/).to_stdout
+        end
+
+        it 'prefers the default over the name' do
+          validation = 'I_AM_THE_DEFAULT'
+          command = {:name => "echo I_AM_THE_NAME", :add_env => {:name => 'FOOBAR', :default => "echo #{validation}"}}
+          task.add_command(command)
+          expect{ described_run_task }.to output(/#{validation}/).to_stdout
+        end
```
er0ck, line 262: i don't think this belongs in rototiller_task_spec.  should be in env specs or possibly command, if env specs don't have a viable parent.  although i guess if there is no parent in env specs, here is as good a place as command.  🍧

docs/rototiller_task_reference.md
```diff
@@ -23,6 +23,9 @@
 <a name="rototiller_task:add_env"></a>
 ### #add_env
 * adds an arbitrary environment variable for use in the task
+* If the parent does call the `name=` method and the method `default=` is not called under `add_env` the value passed to `name=` is the default
```
er0ck, line 27: i think we need a short description near here of what "parent" means in this context.  
'object on which we're calling the add_env method' and not say a parent task

docs/rototiller_task_reference.md
```diff
@@ -159,3 +162,112 @@ produces:
     command_name --switch --option opt arg arguments go last
 
     # what do you think ARG_OVERRIDE2 does?
+
+##Relationship between add_env and parameter name
+Any parameter or task can call the `add_env` method.  The context of the `add_env` method call and additional method calls determine its behavior.  The only location where `add_env` is not available is under an existing `add_env` method call.
```
er0ck: this could use more on what the relationship is, and how it determine's its behavior.  or maybe just "see below".  but here might say something like "environment variable default values are different for tasks and command and its parameters"  "a task's environment variable does not have a parent with a useful name, so it does not set its default"

docs/rototiller_task_reference.md
```diff
@@ -159,3 +162,112 @@ produces:
...
+&nbsp;
+
+    desc "a command that will stop"
+    rototiller_task :stop do |task|
+      task.add_command do |cmd|
+        cmd.name = 'echo command_name'
+        cmd.add_option do |o|
+          o.name = '--option'
+          o.add_argument do |arg|
+            arg.add_env({ :name => 'ARG_OVERRIDE', :message => 'message at the env for argument' })
```
er0ck: we need an explanatory reminder of why this stops.  "arg does not have a name, so its environment variable does not automatically set a default value"

docs/rototiller_task_reference.md
```diff
@@ -159,3 +162,112 @@ produces:
...
+
+
+Stopping and asking for an Environment variable that is not related to a command
+&nbsp;
+
+    desc "a command that will stop"
+    rototiller_task :stop do |task|
+      task.add_command do |cmd|
+      
+        cmd.add_env({ :name => 'ARG_OVERRIDE' })
```
er0ck: i think you want this at the task level?  won't this get a default of "echo command_name"?

er0ck: or does it stop because cmd's name is set below?  that's confusing

docs/rototiller_task_reference.md
```diff
@@ -159,3 +162,112 @@ produces:
...
+      
+        cmd.name = 'echo command_name'
+        cmd.add_option do |o|
+          o.name = '--option'
+          o.add_argument do |arg|
+            arg.name = 'option_argument'
+          end
+        end
```
er0ck: why is this option here?  it just makes the example unclear

docs/rototiller_task_reference.md
zreichert: That was a mistake, moved to the command level.

docs/rototiller_task_reference.md
zreichert: removed

zreichert:

zreichert:

zreichert:

zreichert:

PR #68, opened 2016-11-30
(maint) fixup doc problems

er0ck:
* old docs from pre-1.0 left hanging around, now removed
* remove spurious process for creating yard server
  * point to rubydoc.org instead
* fix reference doc table in GH markdown
* fix reference doc TOC links


zreichert:
👍

er0ck:
so... push the green button?

PR #69, opened 2017-03-28
(QA-2887) Rototiller should not depend on Rake 11

er0ck:
* previously the gemfile would require ~> 11.0 if no env was set
* this causes problems with beaker who requires ~>10.0
* set to anything greater/equal than 0.9.0 unless the env is used


samwoods1:
👍

er0ck:
do we need two signoffs on these?  thoughts?

PR #70, opened 2017-03-29
DO NOT MERGE (QA-2851) Pry not functional during Rototiller task

er0ck:
* open3, Kernel#`, et al do not allow interaction with the subprocess
* this prevents pry and the like in the subprocess started from rototiller
* we change to using Kerne#system here
  ** that results in no output stream for saving
  ** we do still have pid and exit_code
* this is a breaking change due to the output stream issue
* we should look into other methods using pipes to try and save the output stream

this needs more work, i'm hoping sam comes up with a solution using pipes that allows capture of the output streams
and it's blocked as we have to update master's version string to 2.0.0

### DIFF
--- a/lib/rototiller/task/params/command.rb
+++ b/lib/rototiller/task/params/command.rb
@@ -159,22 +159,24 @@ module Rototiller
       def run
         # make this look a bit like beaker's result class
         #   we may have to convert this to a class if it gets complex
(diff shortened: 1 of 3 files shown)


er0ck:
going a different way with Process.spawn and IO pipes

PR #72, opened 2017-04-03
(maint) fix acceptance bundle

er0ck:


er0ck:

acceptance/pre-suite/01_install_rototiller.rb
```diff
@@ -8,8 +8,9 @@
   end
   scp_to(sut, gem_name, gem_name)
 
+  rake_version = `rake --version`.split[2]
   if ENV['RAKE_VER']
```
samwoods1: It looks like you're only relying on the 'rake --version' command now... should we remove this if statement?

er0ck:

acceptance/pre-suite/01_install_rototiller.rb
er0ck: good point.  but i think we want that there so it'll just install whatever version of rake on the SUT if we aren't using rake at all, locally, for instance.

er0ck: it should probably check `if $(which rake)` or similar, but i dunno if i care enough to fix that for now.

er0ck:
so this is failing ubuntu16 against rake 0.9 only.  this is expected and will be remedied when this is merged: https://github.com/puppetlabs/ci-job-configs/pull/2651
(no need to test against 0.9)
@samwoods1 this is ready to go.

PR #73, opened 2017-04-11
(qa-2851) pry not functional during rototiller task

er0ck:


er0ck:

er0ck:

er0ck:
@samwoods1 this is ready to go.  be harsh on the review of this one.  underlying task command run process changed quite a bit.

lib/rototiller/task/params/command.rb
```diff
@@ -161,15 +160,37 @@ def run
...
+          @result.exit_code = @exitstatus
+          write_pipe.close
+        end
+
+        #FIXME: monitor for deadlock?
+        while @exitstatus == :not_done
+          # could probably optimize a bit here by checking the pipe before writing/sleeping?
+          sleep 0.001
+          begin
+            this_read = read_pipe.readline
```
samwoods1: I feel like this would only read a single line, even if more than one line were written during that 0.001 second. It looks like you could use gets(nil) http://ruby-doc.org/core-2.0.0/IO.html#method-i-gets to get everything that has been written since the last time you did a read.

samwoods1 (changes requested):

lib/rototiller/task/params/command.rb
er0ck: only it doesn't. it works.

e.g. i've tested it with commands like:
```
for i in {1..6600}; do echo "$i"; done
```

(which is larger than the default buffer size of ruby pipes in 64bit OS

er0ck: here is one where i just make the command irb, and issue random ruby interactively:
```
[root@m6yqsxhfaxybktf ~]# rake

irb(main):003:0> methods
=> [:to_s, :inspect, :context, :conf, :irb_quit, :exit, :quit, :irb_print_working_workspace, :irb_cwws, :irb_pwws, :cwws, :pwws, :irb_current_working_binding, :irb_print_working_binding, :irb_cwb, :irb_pwb, :irb_chws, :irb_cws, :chws,
:cws, :irb_change_binding, :irb_cb, :cb, :workspaces, :irb_bindings, :bindings, :irb_pushws, :pushws, :irb_push_binding, :irb_pushb, :pushb, :irb_popws, :popws, :irb_pop_binding, :irb_popb, :popb, :source, :jobs, :fg, :kill, :help, :ir
b_exit, :irb_context, :install_alias_method, :irb_current_working_workspace, :irb_change_workspace, :irb_workspaces, :irb_push_workspace, :irb_pop_workspace, :irb_load, :irb_require, :irb_source, :irb, :irb_jobs, :irb_fg, :irb_kill, :i
rb_help, :nil?, :===, :=~, :!~, :eql?, :hash, :<=>, :class, :singleton_class, :clone, :dup, :taint, :tainted?, :untaint, :untrust, :untrusted?, :trust, :freeze, :frozen?, :methods, :singleton_methods, :protected_methods, :private_metho
ds, :public_methods, :instance_variables, :instance_variable_get, :instance_variable_set, :instance_variable_defined?, :remove_instance_variable, :instance_of?, :kind_of?, :is_a?, :tap, :send, :public_send, :respond_to?, :extend, :disp
lay, :method, :public_method, :define_singleton_method, :object_id, :to_enum, :enum_for, :==, :equal?, :!, :!=, :instance_eval, :instance_exec, :__send__, :__id__]
irb(main):004:0> irb_print_working_workspace
=> main
irb(main):005:0> to_s
=> "main"
irb(main):006:0> equal?
ArgumentError: wrong number of arguments (0 for 1)
        from (irb):6:in `equal?'
        from (irb):6
        from /usr/bin/irb:12:in `<main>'
irb(main):007:0> Readline::VERSION
=> "6.2"
irb(main):008:0> quit
[root@m6yqsxhfaxybktf ~]# cat Rakefile
require 'rototiller'
rototiller_task :default do |t|
  t.add_command({:name => 'irb'})
end
```

samwoods1: I think if you measure how long it takes, it would take (6600 * 0.001) seconds to complete since it waits 0.001 seconds to read the next line even if multiple lines have been written, whereas it should complete faster if you use readline(nil) instead of just readline.  It looks like readline takes the same sep parameter, so doesn't need to be gets.

samwoods1: Just played around with it... if you increase the amount of time you wait per iteration, you can see that it definitely is slowing things down.  Turns out that 'readline(nil)' does not work how I understood either though, but 'readpartial(10240)' does work how I thought readline(nil) should work.

er0ck: i agree if you reduce how often it reads from the pipe it will slow it down.  but i'm not sure what the point here is.  do you want it to read from the pipe as often as possible?

samwoods1: read_pipe.readpartial(65536)

er0ck:

er0ck:
@zreichert i think this is good.  take a look if you want input.

@ThoughtCrhyme i believe this addresses your issues.  please try it out if you get a chance.

samwoods1 (approved):

lib/rototiller/task/params/command.rb
```diff
@@ -161,15 +160,40 @@ def run
...
-        }
+
+        read_pipe, write_pipe = IO.pipe
+        begin
+          @result.pid = Process.spawn(self.to_str, :out => write_pipe, :err => write_pipe)
+        rescue Errno::ENOENT => e
+          $stderr.puts e
+          @result.output << e.to_s
+          @result.exit_code = 127
+          raise
```
zreichert, line 171: Should this raise have a message?  I believe this will result in an empty RuntimeError.

er0ck: no.  it re-raises the same exception with the default message.  but i've handled all our result issues in the handler here.
you can see, because this is caught in the unit tests:
https://github.com/puppetlabs/rototiller/pull/73/files#diff-02ef5a5819402594008a7cb344cd6509R61

er0ck:
@zreichert @samwoods1 there are still a few issues with this depending on the version of Readline.  often you can't see the pry prompt at all.  sometimes there is a newline between every character you type (only in display, not according to the pry process).

i think we need to do some more work here, possibly with the bug report email i sent in to Readline...  

on osx specifically, they use Editline and often do not have terminal corruption of characters, but i can't see the pry prompt

same thing when using `irb` as the command in a rototiller_task.
```
mds7ebtug2em5ed:~ root# rake -f /tmp/Rakefile_structure_ctkybfjd allows_downstream_pry



ls
NameError:
undefined local variable or method `ls' for main:Object
        from (irb):2
        from /usr/bin/irb:12:in `<main>'
RUBY_VERSION
=> "2.0.0"
```

PR #77, opened 2017-06-21
(qa-3114) update Rakefile to rototiller 1.y

er0ck:
(QA-3114) update our Rakefile to rototiller 1.0+

    (maint) add test namespace to Rakefile, clarify descriptions of tasks

    * remove descriptions of tasks that aren't useful to users so they don't
    show up in -T
    * add test namespace, rename spec task to test:unit
    * add acceptance task description to remove rototiller default

    (maint) describe rake tasks with -T as default task

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,80 +6,84 @@ task :default do
   sh %{rake -T}
 end
 
-desc "Run spec tests"
-RSpec::Core::RakeTask.new(:test) do |t|
-  t.rspec_opts = ['--color']
-  t.pattern = ENV['SPEC_PATTERN']
+namespace :test do
+
+  desc "Run unit tests"
+  RSpec::Core::RakeTask.new(:unit) do |t|
+    t.rspec_opts = ['--color']
+    t.pattern = ENV['SPEC_PATTERN']
+  end
+
+  task :test => [:check_test]
+
+  task :generate_host_config do |t, args|
+
+    target = ENV["LAYOUT"] || ENV["TEST_TARGET"] || 'centos7-64'
+    generate = "beaker-hostgenerator"
+    generate += " #{target}"
+    generate += " > acceptance/hosts.cfg"
+    sh generate
+    sh "cat acceptance/hosts.cfg"
+  end
+
+  default_rake_ver = '11.0'
+
+  desc "Run acceptance tests"
+  rototiller_task :acceptance => [:generate_host_config] do |t|
+    t.add_env({:name => 'LAYOUT',   :default => 'centos7-64',
+               :message => 'The argument to pass to beaker-hostgenerator',
+               :set_env => true})
+    t.add_env({:name => 'RAKE_VER', :default => default_rake_ver,
+               :message => 'The rake version to use when running unit and acceptance tests',
+               :set_env => true})
+
+    t.add_flag do |flag|
+      flag.name = '--hosts'
+      flag.default = 'acceptance/hosts.cfg'
+      flag.message = 'The configuration file that Beaker will use'
+      flag.override_env = 'BEAKER_HOSTS'
+    end
+    t.add_flag do |flag|
+      flag.name = '--preserve-hosts'
+      flag.default = 'onfail'
+      flag.message = 'The beaker setting to preserve a provisioned host'
+      flag.override_env = 'BEAKER_PRESERVE_HOSTS'
+    end
+    t.add_flag do |flag|
+      flag.name = '--keyfile'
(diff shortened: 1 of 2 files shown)


er0ck:
this probably needs updating to enable migration to ci.next.

@smcelmurry, thoughts?  i forget what tasks we need to enable RMM/ABS

junebug-spencer:
@er0ck Yeah the bhg string will need an optional flag for the abs hypervisor (`--hypervisor abs`). You can probably copy what is being done over in [pe_acceptance_tests](https://github.com/puppetlabs/pe_acceptance_tests/blob/2017.3.x/acceptance/high_availability/rakefile.rb#L29).

You'll also need to to add beaker-abs to the Gemfile:
`gem "beaker-abs", *location_for(ENV['BEAKER_ABS_VERSION'] || "~> 0.2")`

Thanks for checking in with this!

er0ck:
thanks @smcelmurry.

in the pe_acceptance_tests HA setup they use `BEAKER_ABS`
is that typical?  i can't see any jobs that use this variable.
what is the typical value?

i'm realizing rototiller has trouble with 'boolean' env vars, especially when it's a default false/empty

samwoods1:
@er0ck BEAKER_ABS was the ENV var that @sbeaulie suggested and said he would ensure would be there in the ci-next job once HA got migrated.

er0ck:
thanks @samwoods1 note the way we're using it in the Rakefile in this PR.  it's not a great use case for rototiller.  i think i've worked around it ok.  it's just nasty because of limitations in rototiller.  
if we think having default-false boolean env vars like this is the right interface, we'll just have to fix it in future rototiller to make our syntax more clear.

Rakefile
```diff
@@ -2,82 +2,90 @@ require 'rspec/core/rake_task'
...
+      bhg_command.name = "beaker-hostgenerator"
+      bhg_command.add_argument do |argument|
+        # ugh, this is terrible, but it's the only way in current rototiller
+        argument.name = ENV['BEAKER_ABS'] ? ' --hypervisor=abs' : ''
+        argument.message = 'Use hypbervisor abs option for bhg'
+      end
+      bhg_command.add_argument do |arg|
+        arg.name = 'centos7-64'
+        arg.add_env({:name => 'LAYOUT', :message => 'The beaker-hostgenerator pattern'})
+        arg.add_env({:name => 'TEST_TARGET', :message => 'The beaker-hostgenerator pattern'})
```
samwoods1: What is the difference between LAYOUT and TEST_TARGET? Should they have unique descriptions?

er0ck: no difference, LAYOUT is historical, but not very descriptive.  the PA components have moved to TEST_TARGET, or similar

er0ck: i suppose we could state that TEST_TARGET value is used over LAYOUT, if they are both present.  (rototiller defaults to `last`)

Rakefile
```diff
@@ -2,82 +2,90 @@ require 'rspec/core/rake_task'
...
-    flag.override_env = 'BEAKER_TESTS'
+
+  desc ""
+  #FIXME: this is probably a build task, given that it has an output file
+  rototiller_task :generate_host_config do |t|
+    t.add_command do |bhg_command|
+      bhg_command.name = "beaker-hostgenerator"
+      bhg_command.add_argument do |argument|
+        # ugh, this is terrible, but it's the only way in current rototiller
+        argument.name = ENV['BEAKER_ABS'] ? ' --hypervisor=abs' : ''
```
samwoods1, line 38: Yeah... if the env var is a string is't not bad, but if it's a true/false value that's translating into a string or no string, this seems like the best way given current implementation.

samwoods1:
👍

er0ck:
@puppetlabs-jenkins retest me please
just to make ci green because [skip ci] doesn't work on scratch PRs??

er0ck:

er0ck:

er0ck:
if we go with this task name, we have to update the `command` in ci-job-configs

er0ck:
#82 is partly superseded by this.  

I'll fix this one up, get it in, then we can rebase 82

er0ck:
@samwoods1 @zreichert can one of you push go on this, if it looks good?  i've added backwards compat task names that surprisingly work (you can have a task named the same thing as a namespace 😢 ).  

This is gating 82 and the move to ci.next, gating addition of periodic runs, move to travis, etc

edit: sam gave the 👍 on this.  i'll give it a day and then hit go if no one else does

PR #78, opened 2017-06-21
(do not merge) (qa-3110) spike rototiller create integration suite

er0ck:
(QA-3110) spike: rototiller create integration suite

    * implement an integration test helper set
      * this creates a helper that can run a single task in rspec
    * we could create another helper for rake -T and rake -D, etc, maybe
      * or it might make sense to just keep testing that stuff in acceptance
    * this doesn't work
      * it appears to not run the commands, not clear why

    (QA-3110) update CONTRIBUTING doc to reflect new test layout

* this part needs a bit more work on how to run units vs. integration

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,11 +8,11 @@
 ## Making Changes
 
 * Create a topic branch from where you want to base your work.
-  * This is the `master` branch in the case of rototiller
-  * To quickly create a topic branch based on master use `git checkout -b my_contribution master`. Do not work directly on the `master` branch.
+  * This is the `master` branch for breaking changes in major version releases, the `stable` branch for non-breaking changes and bug-fixes in Y or Z releases.
+  * To quickly create a topic branch based on master use `git checkout -b my_contribution master`. Do not work directly on the `master` nor `stable` branches.
 * Make commits of logical _working_ and _functional_ units.
 * Check for unnecessary whitespace with `git diff --check` before committing.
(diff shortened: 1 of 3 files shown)


er0ck:
can you guys figure out why this isn't working?
@zreichert @samwoods1 

ah, right.  Rake::TaskLib only registers a task.  we'll have to reach into rake to run it.

edit:
okay, so rake executes (some of) the ruby in a Rakefile upon load:
https://github.com/ruby/rake/blob/master/lib/rake/application.rb#L767-L770

i tried skipping that step and defining a task in Rake::Application like:
`Rake::Application.new.define_task(Rake::Task, *args, &block)`  
but then it doesn't have `#add_env`, `#add_command`, etc,  in rake::task, of course

or it doesn't have Task methods, because RototillerTask is a child of TaskLib, not Rake::Task 😭 
even if you give it a RototillerTask as scope:
```
[12] pry(#<RSpec::ExampleGroups::RototillerTestTask>)> app = Rake::Application.new.define_task(Rototiller::Task::RototillerTask, 'abstract-me' , &block)
NoMethodError: undefined method `scope_name' for Rototiller::Task::RototillerTask:Class
from /Users/erict/projects/rake/lib/rake/task_manager.rb:33:in `define_task'
```

so to get this to work we might be able to monkey patch `Application.#load_imports` in the test helper here.  not sure how much time i want to spend on this, though.

er0ck:
crap, i always forget about [skip ci] for spikes

PR #80, opened 2017-06-22
(do not merge) (qa-3110) spike rototiller create integration suite2

er0ck:
@samwoods1 @zreichert that's the last i'll probably spend on this effort.  totally stuck and frustrated.  

rake is strange.  or the way we interface with it is.  
i can't get this to put anything on stdout.  

this theoretically should be the same as previous attempt, i just re-implemented most of `RototillerTask`

if you feel like looking at this and helping me figure out WHY it doesn't work, that'd be awesome.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,11 +8,11 @@
 ## Making Changes
 
 * Create a topic branch from where you want to base your work.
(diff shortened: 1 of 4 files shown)


er0ck:
ok, so skip ci didn't work here
fun

PR #81, opened 2017-07-17
(maint) Print command and print message and status the same everywhere

samwoods1:


lib/rototiller/task/params/command.rb
```diff
@@ -161,6 +161,9 @@ def run
         @result = Result.new
         @result.output = ''
 
+        # Print out the command we're running so users can see what was executed
+        puts self.to_str
```
zreichert: Wow were we not doing this?!?

samwoods1: Nope, so you never see the command that is executed anywhere

samwoods1:

er0ck:
i'm okay saying this is not breaking, and putting this in stable.

i suppose changing our output can be viewed as part of our API, but i doubt it's being used that way now...

but i'd like a ticket for this type of change, @samwoods1

lib/rototiller/task/params/env_var.rb
```diff
@@ -47,16 +47,16 @@ def message
         this_message = String.new
 
         if env_status    == STATUS[:nodefault_noexist]
-          this_message << red_text('ERROR: environment-variable not set and no default provided: ')
+          this_message << red_text(' - ERROR: environment-variable not set and no default provided: ')
```
er0ck: why?  i'm not saying i disagree.  just wondering what the thought is for the hyphen

spec/rototiller/task/params/command_spec.rb
```diff
@@ -63,7 +63,7 @@ module Task
         end
         context 'with a block' do
           it 'runs the block' do
-            expect{ command.run { |result| puts "my exit_code: '#{result.exit_code}'" } }.to output("#{@arg_name}\nmy exit_code: '0'\n").to_stdout
+            expect{ command.run { |result| puts "my exit_code: '#{result.exit_code}'" } }.to output(/#{@arg_name}\nmy exit_code: '0'\n/).to_stdout
```
er0ck: why?  there aren't any regex patterns in there

er0ck:
@samwoods1 as you stated in hipchat: i think this needs tests

wtf, github, why can't i make this comment a review?

lib/rototiller/task/params/env_var.rb
samwoods1: Just to give some separation between the message and the status message.  Right now it prints out like this:
`Beaker pre-suiteINFO: no system environment-variable value, using default provided: 'BEAKER_PRE_SUITE': '../../setup/high_availability/install.rb': ''`

This would make it print out like this: 
`Beaker pre-suite - INFO: no system environment-variable value, using default provided: 'BEAKER_PRE_SUITE': '../../setup/high_availability/install.rb': ''`

It could be a space if we just want that.

er0ck: does it really print `Beaker pre-suite` like that?  is that the task name?

spec/rototiller/task/params/command_spec.rb
samwoods1: It used to just print out a message because the command is an echo command.  Now it prints the entire command, and then the echo'd message.  I thought about using .end_with? but that doesn't work with the output method.  I guess I could change the regex to ensure it's at the end.

er0ck: ah.  yeah, the `$` at end of pattern would make it more clear.

samwoods1:
@er0ck @zreichert I made a significant change to this, can you re-review?  I reverted the changes to add the ' - ' to the status messages.  Basically, the way it used to be, the messages for environment variables were printed out one way:
`INFO: environment-variable overridden from system, not using default: 'pe_family': default: '2016.5' using: '2017.3': 'The PE Family to install on hosts'`

And messages for options/switches were printed another way (since they have an environment variable under the option, and not directly under a command):
`The log level under which you want beaker to runINFO: environment-variable overridden from system, not using default: 'BEAKER_LOG_LEVEL': default: 'debug' using: 'debug': ''`

I changed it to always be the status message first, followed by the user defined message, followed by a line break.

samwoods1:

er0ck:
that part looks good to me.  nice work!

still needs tests for printing the command

lib/rototiller/task/params/command.rb
er0ck: We do this with {{--verbose}} turned on in the Command loop in rototiller_task:
https://github.com/puppetlabs/rototiller/blob/master/lib/rototiller/task/rototiller_task.rb#L108
if we think we want this all the time, we can remove the conditional.  we should probably change its color, or make it clear we're logging it, and it's not the shell.

er0ck:
@samwoods1 this needs a rebase, and has some comments and updates required.

also, maybe you could apply this change to https://tickets.puppetlabs.com/browse/QA-2802, and/or https://tickets.puppetlabs.com/browse/QA-2754

i think this is too large of a change for (maint)

er0ck:
@samwoods1 rebase this please, see comments above as well.  so close!

PR #83, opened 2017-08-25
(maint) fix bhg task in Rakefile against upstream changes

er0ck:
* ensure there is not an old hosts file lying around
* actually redirect output to the hosts file O:)


junebug-spencer (approved):
Needed for CI.Next in puppetlabs/ci-job-configs#3373

PR #86, opened 2018-03-08
(maint) improve code quality rake tasks

er0ck:
i thought we could avoid the bundle install on the host (previous PR).  but we now use a bunch of gem's provided rake tasks in the rakefile.  so we need the gem or the bundle.  installing the gems in the right version against the host ruby version is a PITA.  
we need to unit test against a couple ruby versions anyway, we've had issues there.  so i'll switch back to rvm.  i'll try to narrow it to versions where binaries are available.  i don't really know what platform the host vm is on.
the bundle on the host tends to pollute the bundle in the container, but i believe i've largely tackled that.

more to come...


Rakefile
```diff
@@ -1,29 +1,122 @@
...
+
+# rubocop:disable Metrics/BlockLength
+namespace :docs do
+  # docs:yard task
+  YARD::Rake::YardocTask.new
+
+  desc "Clean/remove the generated YARD Documentation cache"
+  task :clean do
+    original_dir = Dir.pwd
+    Dir.chdir(File.expand_path(File.dirname(__FILE__)))
```
samwoods1: Why are we changing directories? We can remove a directory without cd'ing into it.

Rakefile
```diff
@@ -1,29 +1,122 @@
...
+    sh "rm -rf #{YARD_DIR}"
```
samwoods1: Why use shell? We should use FileUtils.rm_rf

Rakefile
```diff
@@ -1,29 +1,122 @@
...
+    verify.threshold = 57 # FIXME
+  end
+
+  desc "Generate static project architecture graph. (Calls docs:yard)"
+  # this calls `yard graph` so we can"t use the yardoc tasks like above
+  #   We could create a YARD:CLI:Graph object.
+  #   But we still have to send the output to the graphviz processor, etc.
+  task arch: [:yard] do
+    original_dir = Dir.pwd
+    Dir.chdir(File.expand_path(File.dirname(__FILE__)))
```
samwoods1: Again, not sure why we need to chdir

Rakefile
er0ck: randell asked the same thing.  i need to comment this, because i keep forgetting myself.  i think the idea is that rake can be run from any dir below project root, so we never know what dir we're in.  but we know _where_ the rakefile is with `__FILE__`.  but i guess we don't have to cd to the dir.  i can just absolute path the yarddir with the __FILE__ path.

Rakefile
er0ck: in this case we might have to `chdir` because yard needs to work on code in a relative path.. and according to the `yard graph` manpage you can't specify a path to analyze.

er0ck:
@samwoods1 updated

Rakefile
```diff
@@ -1,29 +1,124 @@
...
+    graph_processor = "dot"
+    if exe_exists?(graph_processor)
+      FileUtils.mkdir_p(DOCS_DIR)
+      if system("yard graph --full | #{graph_processor} -Tpng " \
+          "-o #{DOCS_DIR}/arch_graph.png")
+        puts "we made you a class diagram: #{DOCS_DIR}/arch_graph.png"
+      end
+    else
+      puts "ERROR: you don't have dot/graphviz; punting"
```
RandellP: Is it really an error if it doesn't cause a failure.  Should it be a warning?

Rakefile
```diff
@@ -1,29 +1,124 @@
...
+  RubyCritic::RakeTask.new
+  RubyCritic::RakeTask.new do |task|
+    task.paths   = FileList['lib/**/*.rb']
+  end
+
+  begin
+    require "rspec/core/rake_task"
+    RSpec::Core::RakeTask.new(:spec)
+    desc "Run unit tests"
+    RSpec::Core::RakeTask.new do |t|
```
RandellP: What does this second new do?  What task does it create?

Rakefile
er0ck: it fails to produce the image.  because it can't convert the dot format to png.
i could put that in stderr, if that helps, and exit 1

Rakefile
er0ck: actually, i think the first one is a duplication bug.  or really does nothing and is overwritten by the second which has some other config args.  i think i stripped this out in the gem_of version

Rakefile
```diff
@@ -1,29 +1,123 @@
...
+    Dir.chdir(File.expand_path(File.dirname(__FILE__)))
+    graph_processor = "dot"
+    if exe_exists?(graph_processor)
+      FileUtils.mkdir_p(DOCS_DIR)
+      if system("yard graph --full | #{graph_processor} -Tpng " \
+          "-o #{DOCS_DIR}/arch_graph.png")
+        puts "we made you a class diagram: #{DOCS_DIR}/arch_graph.png"
+      end
+    else
+      STDERR.puts "ERROR: you don't have dot/graphviz; punting"
```
RandellP: To me the "punting" implied it was okay for it to fail, and things shouldn't stop.    The "ERROR" conflicted with that.  For me, I don't think generating the class diagram is really critical.  I don't think it gets checked in or anything.  So I would have gone the warning route.  But up to you.

Rakefile
```diff
@@ -1,29 +1,123 @@
...
+  FlayTask.new :flay, allowed_repitition, %w[lib]
+  RoodiTask.new
+  RubyCritic::RakeTask.new do |task|
+    task.paths   = FileList['lib/**/*.rb']
+  end
+
+  begin
+    require "rspec/core/rake_task"
+    desc "Run unit tests"
+    RSpec::Core::RakeTask.new do |t|
```
RandellP: Wouldn't hurt to comment on what the task name is here... or make it explicit in another way...

er0ck:
@samwoods1 @RandellP this is ready to go

RandellP (approved):

PR #89, opened 2018-04-20
(slv-96) rototiller run acceptance on test runner, add distelli build step

er0ck:


er0ck:
@tvpartytonight @kevpl, FYI  kinda interesting beaker use here.

Dockerfile-tests
```diff
@@ -0,0 +1,46 @@
...
+  ruby-devel \
+  zlib-devel \
+  openssh-server \
+  which \
+  && yum clean all
+
+# install RVM, Ruby, and Bundler
+RUN gpg --keyserver hkp://keys.gnupg.net --recv-keys D39DC0E3
+RUN \curl -L https://get.rvm.io | bash -s stable
+RUN /bin/bash -l -c "rvm install 2.4"
```
RandellP: I assume you are doing the bash to get the rvm.  Couldn't you just run the source command that it added to your rc?  That would avoid spawning several additional shells which would make it a little faster I think.

Rakefile
```diff
@@ -155,6 +156,17 @@ namespace :docs do
...
 end
 
+namespace :container do
+  desc "(re)build docker container for tests"
+  rototiller_task :build do |command|
+    command.name = "docker build ./ --file Dockerfile-tests --tag pcr-internal.puppet.net/slv/rototiller:latest"
+  end
+  desc "push docker container for tests to Puppet Container Registry. You need to be logged into it first `docker login -u TOKEN --password <token from pcr> pcr-internal.puppet.net`"
+  rototiller_task :push do |command|
+    command.name =  "docker push pcr-internal.puppet.net/slv/rototiller:latest"
```
RandellP: the pcr uri should probably be in a varaible near the top or something, since it is used more than once, and highly specific.

distelli-manifest.yml
```diff
@@ -18,3 +21,4 @@ slv/rototiller:
     - rvm use 2.4.0
     - bundle install --without system_tests
     - bundle exec rake test:unit
+    - docker run --interactive --env CI --volume $(pwd):/rototiller pcr-internal.puppet.net/slv/rototiller:latest /bin/bash -l -c " /usr/sbin/sshd && bundle update && bundle exec beaker --debug --no-ntp --repo-proxy --no-validate --no-provision --keyfile /root/.ssh/docker_acceptance --load-path acceptance/lib --pre-suite acceptance/pre-suite --hosts acceptance/hosts.cfg --tests acceptance/tests "
```
RandellP: Isn't this what the rake task you added creates?  Couldn't you use the rake task here?

RandellP (changes requested):

distelli-manifest.yml
er0ck: this was a proof of concept.  the rake task, i used prior.  but it wasn't giving me any debug output for why it wasn't working.  this got me closer.  turns out they don't allow mounting local FS volume in the container.  so this might not work at all. 

i'll probably have to copy the working dir into the container (or just git clone), and then fetch/reset whenever we run the container to get the PR

Dockerfile-tests
er0ck: i'm not sure why, but if you don't say /bin/bash here, it can't find rvm. 
are you saying we can just source the rc file it included?
`source /usr/local/rvm/scripts/rvm`

er0ck: hmmm, that script doesn't exist.

er0ck:
@RandellP this works now and is ready to merge, if you are

Rakefile
```diff
@@ -155,6 +186,47 @@ namespace :docs do
   end
 end
 
+namespace :container do
+  desc "(re)build docker container for tests"
+  rototiller_task :build do |t|
+    t.add_command do |command|
+      #command.name = "docker build ./ --file Dockerfile-tests --tag"
+      # WARNING: this will delete any .bundle and Gemfile.lock
+      command.name = "rm -rf Gemfile.lock .bundle/ && docker build ./ --file Dockerfile-tests --tag"
```
RandellP: Why the rm -rf stuff.  Assuming there is a good reason... :)  Add a comment about why.

RandellP (changes requested):
Feels like there should be some kind of readme update with some info on all this container stuff.

Rakefile
er0ck: ah.  actually, i can probably remove that now.  forgot it was there.  there was a time when there was a local bundle and a container bundle and they'd conflict.

RandellP (approved):

PR #90, opened 2018-05-15
(slv-55) rototiller soften language/colors in env var messages

er0ck:
most critical files to review include env_var.rb and rototiller_task.rb where i changed the messaging ordering and wording.  also their related tests


docs/rototiller_task_reference.md
```diff
@@ -53,29 +53,37 @@
...
-      task.add_command({:name => "echo 'i am testing everything with $RANDOM_VAR = #{ENV['RANDOM_VAR']}'"})
-      task.add_command({:name => "echo 'some other command!'"})
+      # most methods take either a hash, or block syntax (see next task)
+      ENV['HAS_VALUE_NO_DEFAULT'] = 'from environment'
+      task.add_env({:name     => 'HAS_VALUE_NO_DEFAULT'})
+      # rototiller will set this in the environment so the task, programs, can use it
+      task.add_env({:name     => 'NO_VALUE_HAS_DEFAULT',  :default => 'default value'})
+      ENV['HAS_VALUE_HAS_DEFAULT'] = 'from environment'
+      task.add_env({:name     => 'HAS_VALUE_HAS_DEFAULT', :default => 'default value'})
+      #task.add_command({:name => 'echo NO_VALUE_NO_DEFAULT:  "$NO_VALUE_NO_DEFAULT"'})
```
samwoods1: Why is this line commented out?

lib/rototiller/task/params/env_var.rb
```diff
@@ -44,21 +44,25 @@ def initialize(args={}, &block)
       end
 
       # The formatted messages about this EnvVar's status to be displayed to the user
+      # @param indent [String] how far to indent each message
       # @return [String] the EnvVar's message, formatted for color and meaningful to the state of the EnvVar
-      def message
+      INDENT_ARRAY = ["","  ","    ","      ","        ","          "]
+      def message(indent=0)
+        # we only need up to indent 4, really
+        indent = 1 if indent > 1
```
samwoods1: Why is this getting reset to one if it is greater than one? Shouldn't it just use the indent passed in?

er0ck: yeah, we don't need this no mo.  i was just leaving it in, in case we did.  it used to be `4 if > 4`

i can remove it

PR #91, opened 2018-08-29
(slv-106) rototiller use gem of for gemfile and rake tasks

er0ck:
yes this is a lot of commits and lines.  sorry. 

you can ignore the rubocop fix commits, unless you want to know what we had to go through to refactor the methods (not much other than further abstraction of private methods).


er0ck:
failed due to gem_of.
it's probably not pulling the submodule

er0ck:
ugh.. fixed that, but we don't have the bundle installed on the test-runner, just in the container.  

so i guess i have to figure out if we want docs and lint to run in the container also, or not.

PR #92, opened 2018-08-30
(slv-57) rototiller ensure/test can use methods on a previously created param object

er0ck:


CONTRIBUTING.md
```diff
@@ -40,6 +41,7 @@
 
 ## Testing
 
+* (!) ensure you have gem_of installed in your gems. Bundler can not install it, and our Gemfile requires it.
```
samwoods1: That means bundler won't be able to update it either? How do we know when we need to update the version of gem_of?

lib/rototiller/task/rototiller_task.rb
```diff
@@ -88,7 +97,13 @@ def add_command(*args, &block)
 
       private
 
-      # @private
+      # @api private
+      def print_messages
+        puts @commands.messages
```
samwoods1: Was this supposed to be part of a different commit? It doesn't look like it is part of increasing yard documentation and I don't see any references to it.

spec/rototiller/task/params/env_var_spec.rb
```diff
@@ -74,6 +74,10 @@ module Task
         end
       end
 
+      it "sets its value to parent's name when no default given" do
+        no_name = { name: "my_super_cool_name", parent_name: "parentParamName" }
+        expect(described_class.new(no_name).message).to match(/default: 'parentParamName'/)
```
samwoods1: I am confused by this. It looks like we're setting no_name equal to a hash which has a name value, then checking to make sure default value in the message matches the parentParamName? Shouldn't it just equal the name since it is specified?

lib/rototiller/task/params/command.rb
```diff
@@ -213,7 +215,8 @@ def stop
       # TODO make private method? so that it will throw an error if yielded to?
       def message(indent = 1)
         return_message = ""
-        return_message = "  #{@message}\n" if @message && @message != ""
+        return_message = "  " + green_text("with message: ") +
+          "#{@message}\n" if @message && @message != ""
```
samwoods1: Why is one string inline and the other using +?

CONTRIBUTING.md
er0ck: correct.  submodules are pinned to git shas.  that is then committed to the repo using the submodule.   so... we update it when we want to update 🤷‍♂️

lib/rototiller/task/rototiller_task.rb
er0ck: weird.  it must have been part of making other methods shorter, but then i abandoned it.  i think it can be removed.  i'll do that.

CONTRIBUTING.md
er0ck: oh, also... these commits are part of #91 and just sitting here until it is merged and this one can be rebase.  only the latest 7 commits are applicable here.

lib/rototiller/task/rototiller_task.rb
er0ck: fixed up in #91

lib/rototiller/task/params/command.rb
samwoods1: It is showing this as resolved, but I don't see any change.

er0ck: because it's resolved on #91.  but it's weird that it shows resolved here, as this branch has not been rebased on the branch in #91 🤷‍♂️

er0ck: weird, yeah, i couldn't find this comment except in this email.  i still see this being on the branch

er0ck: so, what's the question?  i added a green text header here so users can differentiate the message from the echo of the command string. 
it's still inline, i just had to call a method and decided to use `+`.  i guess i could have interpolated the method into the string.  it still has the same trailing conditional

samwoods1:
It just seems strange to use interpolation in one place and then use + in
another on the same string.

lib/rototiller/task/params/command.rb
er0ck: ok.  fixed

billclaytor (approved):
Looks good!

PR #94, opened 2018-10-09
(slv-124) ability to redact sensitive env var values in rototiller

er0ck:


samwoods1:
I have a general question.. Do we print out the cmd by default with all of the values? I don't see anywhere we do that, but I know we talked about it.
Right now, to_str or to_s would show the value of the sensitive env_var, which probably wouldn't be desired, BUT we're using to_str to actually construct and execute the command and it has to be the actual value there.  Do we need a different, private method to construct the actual command vs to_str which is public and should redact the sensitive values?

er0ck:
> I have a general question.. Do we print out the cmd by default with all of the values? I don't see anywhere we do that, but I know we talked about it.
> Right now, to_str or to_s would show the value of the sensitive env_var, which probably wouldn't be desired, BUT we're using to_str to actually construct and execute the command and it has to be the actual value there. Do we need a different, private method to construct the actual command vs to_str which is public and should redact the sensitive values?

good question.  maybe. there are lots of ways rake task authors can get around this and print the value including `puts ENV['blah']` so i'm not sure how much i want to protect them from themselves, including restricting `puts env_sensitive`.
also, a `rototiller_task` is not a `RototillerParameter` so it wouldn't have access to a private method (and probably not a protected one).  although i guess a task should never have to print it (nor be able to). which i guess is your point?  hmmm....

also, i don't think i adequately test that a command can still get a sensitive env var's value...  (i'll fix that up).  oh also, the command is printed with all its values, that is probably still broken if you use a sensitive env var to override a command or parameter.

er0ck:
crap.  this makes printing the command before its run kinda difficult.  right now we don't know whether an env_var overrode a parameter name at all, let alone if that override is sensitive.  to complicate this further, we could have any number of env_vars attached to an override.  i can plumb this through, but it'll be more work.

we only just recently added printing the command.  i view it as pretty useful, but i could just remove that feature, i suppose (that'd be breaking)

docs/env_var_example_reference.md
```diff
@@ -42,8 +41,30 @@ Because a task has a name that would not be implicitly useful as the default val
       end
     end
 
+### A task author wants input from the user, but not to print the environment's value
```
billclaytor: This might be more clear as something like: "A task author wants input from the user, but does not want to print the environment variable's value"

er0ck:
> > I have a general question.. Do we print out the cmd by default with all of the values? I don't see anywhere we do that, but I know we talked about it.
> > Right now, to_str or to_s would show the value of the sensitive env_var, which probably wouldn't be desired, BUT we're using to_str to actually construct and execute the command and it has to be the actual value there. Do we need a different, private method to construct the actual command vs to_str which is public and should redact the sensitive values?
> 
> good question. maybe. there are lots of ways rake task authors can get around this and print the value including `puts ENV['blah']` so i'm not sure how much i want to protect them from themselves, including restricting `puts env_sensitive`.
> also, a `rototiller_task` is not a `RototillerParameter` so it wouldn't have access to a private method (and probably not a protected one). although i guess a task should never have to print it (nor be able to). which i guess is your point? hmmm....
> 
> also, i don't think i adequately test that a command can still get a sensitive env var's value... (i'll fix that up). oh also, the command is printed with all its values, that is probably still broken if you use a sensitive env var to override a command or parameter.

ok. i got this portion done and tested.  works well and i think the solution is _decent_

er0ck:
oops. forgot to check rubocop, etc

lib/rototiller/task/params/command.rb
```diff
@@ -165,19 +130,34 @@ def add_argument(*args, &block)
...
+      #   if any params values are set from sensitive env vars, redact them
+      # @api public
+      # @example puts command.safe_print
+      # TODO make private method? so that it will throw an error if yielded to?
+      def safe_print
+        if @is_value_sensitive
+          my_value = "[REDACTED]"
+        elsif name
+          my_value = name
+        end
```
er0ck, line 144: weird that rubocop didn't catch this.  this should probably be assigned based on the output of the conditional, instead of two possible assignments inside the conditional. 🤷‍♂️

